PR #88, opened 2016-04-06
Add pkg to kickstart via API

phracek:

### DIFF
--- a/common.sh
+++ b/common.sh
@@ -23,6 +23,7 @@ KICKSTART_README=$KICKSTART_DIR/README
 COMMON_DIR=$XCCDF_VALUE_REPORT_DIR/common
 DIST_NATIVE=$XCCDF_VALUE_DIST_NATIVE
 DEVEL_MODE=$XCCDF_VALUE_DEVEL_MODE
+SPECIAL_PKG_LIST=$KICKSTART_DIR/special_pkg_list
 
 RESULT_PASS=$XCCDF_RESULT_PASS
 RESULT_FAIL=$XCCDF_RESULT_FAIL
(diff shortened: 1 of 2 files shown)


preup/script_api.py
```diff
@@ -532,6 +532,16 @@ def print_home_dirs(user_name=""):
         pass
 
 
+def add_pkg_to_kickstart(pkg_name):
+    tmp = get_file_content(SPECIAL_PKG_LIST, "rb", True)
+    for pkg in pkg_name.split():
+        tmp.append(pkg)
+
+    write_to_file(SPECIAL_PKG_LIST, "wb", True)
```
pirat89: missing parameter

pirat89:
non functional.
Wouldn't be just append of new pkgs to the end of the file more effective then:
read file -> append pkgs -> write all back to the file?

phracek:
Well, I have appended packages to a file.

phracek:
@pirat89 Can you please review it again? What do you think about it?

preup/script_api.py
```diff
@@ -532,6 +532,12 @@ def print_home_dirs(user_name=""):
         pass
 
 
+def add_pkg_to_kickstart(pkg_name):
+    for pkg in pkg_name.split():
+        write_to_file(SPECIAL_PKG_LIST, "a+b", pkg)
+    return 0
+
+
```
pirat89: what about:
```python
def add_pkg_to_kickstart(pkgs):
    empty = False
    if(!isinstanceof(pkgs, list)):
        # string - pkgs delimited by whitespace
        if(len(pkgs.strip()) == 0):
            empty = True
        else:
           # make list from string
            pkgs = pkgs.strip().split()
    else:
        # list of packages
        if(len(pkgs) == 0):
            empty = True
    if(empty is True):
        log_debug("Missing parameters! Any package will be added.")
        return 1
    write_to_file(SPECIAL_PKG_LIST, "a+b", pkgs)
    return 0
```

It accepts list or string. I count that developer will use more list than strings in this case. Function write_to_file accept list & string too. It's untested, so maybe there are mistakes. Just written here as comment.

common.sh
```diff
@@ -427,6 +428,18 @@ print_home_dirs() {
   conf_get_section "$_uconf_file" "home-dirs"
 }
 
+#Function adds a package to special_pkg_list
+add_pkg_to_kickstart() {
+  [ $# -eq 0  ] && {
+    echo "Missing parameters! Any package will be added." >&2
```
pirat89: May log_debug will be better in this case. Not sure about that. What do You think?
```bash
log_debug "Missing parameters! Any package will be added."
```

PR #97, opened 2016-05-12
API cleanup and documentation.

phracek:

What is changed from man point of view:
- man/preupg.1
- man/preupg-content-creator -- added example from preupg-content-creator, prerequisites
- man/preupgrade-assistant-api.1 -- added all functions which are visible from content point of view.

Script_api changes:
log -> _log all references like log_info uses _log.
log_risk -> _log_risk references like log_high_risk uses _log_risk.
added exit_failed to BASH API.
Functions are better documented. Even log functions.
function switch_to_content -> _switch_to_content
Hide load_pa_configuration->_load_pa_configuration
Hide space_trim -> _space_trim


phracek:
@AloisMahdal @pirat89 Please look on this PR and commented.

man/preupg.1
```diff
@@ -38,9 +48,9 @@ Clean all files created by preupg about previous state of the machine.
 Suppress user interaction.
 .TP
 .B \-\-kickstart
-Generate kickstart.
+Generates kickstart.
 .TP
-.B \-l, --list
+.B \-l, --list-contents-set
```
pirat89: why somewhere is "-" escaped and somewhere not? isn't that mistake?

pirat89:
just quick look, but NACK for me

 - there is missing descriptions/documentation for some variables in man page for API
 - Some variables have comments in API for Python but variables are not commented in API for bash
 - little wierd/random escaping in ma page - see comment above, I am not sure about effect

phracek:
Please be concrete and not abstract. I don't think that first commit will be merge. It will need some iterations of course.
But your comments are relevant. I am going to update them.

pirat89:
It was just quick feedback Petr, I didn't have more time to be more specific. I have looked at common.sh file and it seems almost OK. Found jst issue inside  *add_pkg_to_kickstart* - I pushed commit already.

I was thinking about functions *conf_get_section* and "conf_get_sections". I know that these functions we will need rewrite this cycle because of planned changes in hiearchy/distribution/... of configuration files. So probably these shouldn't be private.

In addition. These labels are too similar and maybe we could change names for better orientation in our API. Like  "get_conf_section" and "get_conf_list_sections". But still, get_conf_section would be private and we would prefer function which return specific value for given variable. It's question, if we should bound it to specific section or not. And what about function which returns just keys from section? get_conf_section would do that.

AloisMahdal:
Hi, sorry for late reply; I was on PTO and then I had other tasks. I'm going through the changes now.

However, I guess my feedback will be more related to API as is, rather than to this particular changeset-- this changeset really just scratches the surface of several things, but some of them may become rather moot after (if ever) my feedback gets implemented.

IOW, I'm more likely to open new issue or ML thread.

common.sh
```diff
@@ -1,31 +1,59 @@
...
+# Directory with scripts which can be executed after installation by administrator
 KICKSTART_SCRIPTS=$KICKSTART_DIR/scripts
+# The same as above
 KICKSTART_POSTUPGRADE=$KICKSTART_SCRIPTS
+# Variable which refers to static data used by preupgrade-assistant and modules
 COMMON_DIR=$XCCDF_VALUE_REPORT_DIR/common
 DIST_NATIVE=$XCCDF_VALUE_DIST_NATIVE
+# Variable which indicates DEVEL mode.
 DEVEL_MODE=$XCCDF_VALUE_DEVEL_MODE
+# Variable which contains file with packages add to the kickstart anyway
```
AloisMahdal: please don't write state the obvious: `Variable which contains` when documenting variable and of course it will contain something ;)

common.sh
```diff
@@ -1,31 +1,59 @@
 
 CACHE=/var/cache/preupgrade
+# Directory with logs gathered by preupgrade-assistant
```
AloisMahdal: this and whole section: it's impossible to know for sure which comments belongs to which variable!

AloisMahdal: it's called `CACHE` and it "contains logs": this is confusing. see also #101

common.sh
```diff
@@ -71,99 +99,165 @@ log()
...
+#informational message
 log_info()
 {
-    log "INFO" "$@"
+    _log "INFO" "$@"
 }
 
+#log_error(message, component=None) -> None
+
+##log message to stdout with severity error
```
AloisMahdal: 2 hash signs?

common.sh
```diff
@@ -71,99 +99,165 @@ log()
         fi
     fi
 
-    echo "$SEVERITY $COMPONENT: $1"
+    echo "$SEVERITY $COMPONENT: $1" >&2
```
AloisMahdal: Are we sure all logs should go to stderr? What goes to stdout? I'm not disagreeing, just curious; this will need to be reflected in our Red Hat internal test API.

AloisMahdal:
By the way, are you using some kind of post-processing (like ReSTructured) for the docs?  Sometimes I see notation like `:param foo`, sometimed `@foo`, I wonder if it's significant

common.sh
```diff
@@ -438,10 +521,10 @@ add_pkg_to_kickstart() {
   }
 
   while [ $# -ne 0 ]; do
-    echo $1 >> SPECIAL_PKG_LIST
+    echo $1 >> "$SPECIAL_PKG_LIST"
```
AloisMahdal: This is a functional change; way outside "api and doc cleanup"

man/preupg.1
```diff
@@ -38,9 +48,9 @@ Clean all files created by preupg about previous state of the machine.
 Suppress user interaction.
 .TP
 .B \-\-kickstart
-Generate kickstart.
+Generates kickstart.
```
AloisMahdal: OK... although the grammatical mood is not really consistent: you changed imperative to indicative here, and added imperative few paragraphs later. It would be nice to have it consistent. (Imperative seems to be more common.)

preup/script_api.py
```diff
@@ -98,26 +98,45 @@
 )
 
 CACHE = "/var/cache/preupgrade"
+# Directory with logs gathered by preupgrade-assistant
```
AloisMahdal: whole section: same issues as with common.sh

man/preupg-content-creator.1
```diff
@@ -6,18 +6,38 @@
...
+\& preupg-content-creator generates these files which should be updated:
+\& - content was created in directory /root/RHEL6_7/database/sqlite.
+\& - ini file which defines a content is RHEL6_7/database/sqlite/sqlite.ini.
+\& - check script which does an assessment is RHEL6_7/database/sqlite/check.sh. Update it before an usage
+\& - solution text which informs about incompatilibies is RHEL6_7/database/sqlite/solution.txt. Update it before an usage
+\&
+\& For testing content run these two commands:
+\& - preupg-create-group-xml RHEL6_7/database/sqlite to create XML file
+\& - preupg -c RHEL6_7-results/database/sqlite/all-xccdf.xml
+\& #
```
AloisMahdal: content or module (or rule, as in `--list-rules`)?  we need to decide and start using only one name instead of creating more and more confusion

phracek: Good point.
Content definitely not.
I would prefer modules. Rules are taken from openscap.
I will modify preupg CLI so it will refer to modules like --list-modules.

common.sh
pirat89: I believe they should. Just example: you create function get_list_.... and uses log_* inside. In that case, you will get some expected output mixed with logs. All logs like log_high_risk, log_warning, log_info, .... should be really on stderr.

AloisMahdal: @pirat89 Makes sense; call me convinced.

common.sh
```diff
@@ -1,31 +1,59 @@
...
 KICKSTART_README=$KICKSTART_DIR/README
+# Directory with scripts which can be executed after installation by administrator
 KICKSTART_SCRIPTS=$KICKSTART_DIR/scripts
+# The same as above
 KICKSTART_POSTUPGRADE=$KICKSTART_SCRIPTS
+# Variable which refers to static data used by preupgrade-assistant and modules
 COMMON_DIR=$XCCDF_VALUE_REPORT_DIR/common
 DIST_NATIVE=$XCCDF_VALUE_DIST_NATIVE
+# Variable which indicates DEVEL mode.
 DEVEL_MODE=$XCCDF_VALUE_DEVEL_MODE
```
AloisMahdal: Should DEVEL_MODE be public at all?

phracek:
I will create separate Pull Requests. One for API and second one for Documentationt stuff.

PR #110, opened 2016-06-14
First api refactoring

phracek:
Signed-off-by: Petr Hracek phracek@redhat.com


common.sh
```diff
@@ -2,18 +2,33 @@
...
+PREUPG_ALLMYFILES=$VALUE_ALLMYFILES
 VALUE_EXECUTABLES=$PREUPGRADE_CACHE/executable.log
+PREUPG_EXECUTABLES=$VALUE_EXECUTABLES
 VALUE_TMP_PREUPGRADE=$XCCDF_VALUE_TMP_PREUPGRADE
+PREUPG_TMP_PREUPGRADE=$VALUE_TMP_PREUPGRADE
+CONFIG_FILES_DIR=$PREUPG_TMP_PREUPGRADE/config_files
+
 POSTUPGRADE_DIR=$VALUE_TMP_PREUPGRADE/postupgrade.d
+PREUPG_POSTUPGRADE_DIR=$POSTUPGRADE_DIR
 CURRENT_DIRECTORY=$XCCDF_VALUE_CURRENT_DIRECTORY
```
AloisMahdal: Why are you duplicating every variable?

To my understanding, aim in longer term was to get rid of them (from the public API).

common.sh
```diff
@@ -160,9 +175,247 @@ exit_fixed()
...
 }
 
+_get_cached_command()
+{
+    # Function gets a cached command
+    # input parameter is command provided by get_cached function
+    command=$1
+    if [ ! -f "$command" ]; then
+        log_error "File $command does not exist. It is mandatory."
+    else
```
AloisMahdal: User did not ask for file but cached command output.  The message should reflect that.

common.sh
```diff
@@ -160,9 +175,247 @@ exit_fixed()
...
+                *)
+                    log_error "Unknown $key for $command. Supported are 'executable', 'allmyfiles', 'rpmtrackedfiles', 'rpmrhsignedfiles'."
+                    return 1
+            esac
+            ;;
+        "info")
+            ;;
+        *)
+            log_error "Unknown command $command. Supported are 'file', 'command', filelist', 'info'."
+            return 1
```
AloisMahdal: This is not true.  'info' is not supported.

I don't think it makes sense to even mention it anywhere.  It just does not exist, OK? ;-)

common.sh
```diff
@@ -160,9 +175,247 @@ exit_fixed()
...
+            _check_config_file "unchecked" "$filename" && return 1
+            ;;
+        "deprecated")
+            _check_config_file "clean" "$filename" \
+            || log_warning "You want to store the file into ${CLEAN_FILES} directory."
+            _check_config_file "dirty" "$filename" \
+            || log_warning "You want to store the file into ${DIRTY_FILES} directory."
+            _check_config_file "unchecked" "$filename" \
+            || log_warning "You want to store the file into ${UNCHECKED_FILES} directory."
+            _check_config_file "deprecated" "$filename" && return 1
```
AloisMahdal: What's the purpose of these checks? I can't decipher that neither from code nor from error messages.

common.sh
```diff
@@ -160,9 +175,247 @@ exit_fixed()
...
+            log_error "Supported options are 'migration', 'upgrade'."
+            ;;
+    esac
+
+}
+
 switch_to_content()
 {
-    cd $CURRENT_DIRECTORY
+    cd ${CURRENT_DIRECTORY}
```
AloisMahdal: I don't see how this improves anything.

Frankly I have no idea what the function is for; the name is confusing (you can't "switch to content", module ir rule).

common.sh
```diff
@@ -160,9 +175,247 @@ exit_fixed()
...
+{
+    # Add filename /root/preupgrade/kickstart directory
+    # File as first parameter has to exists in modules directory
+    # Usage:    add_to_kickstart_dir <filename> <description>
+    script_name=$1
+    kickstart_name="$KICKSTART_DIR/`basename $script_name`"
+    [ ! -f "$script_name" ] && log_error "$script_name does not exist." && exit_error
+    [ -f "$kickstart_name" ] && log_warning "`basename $script_name` already exists in $KICKSTART_DIR"
+    cp -af ${script_name} ${kickstart_name}
+}
```
AloisMahdal: If it adds files, why it's called `add_kickstart_dir`?  Also, can't we add directories?

common.sh
```diff
@@ -160,9 +175,247 @@ exit_fixed()
...
+    esac
+    mkdir -p `dirname ${backup_file}`
+    cp -a ${filename} ${backup_file}
+    return 0
+}
+
+add_kickstart_dir()
+{
+    # Add filename /root/preupgrade/kickstart directory
+    # File as first parameter has to exists in modules directory
```
AloisMahdal: I don't understand this restriction: why the file has to exist in particular directory?  And what is "modules directory" anyway?

IMO only requirement we should make is that the file exists, (i.e. the copy action succeeds).

common.sh
```diff
@@ -160,9 +175,247 @@ exit_fixed()
...
+    # Usage:    add_to_kickstart_dir <filename> <description>
+    script_name=$1
```
AloisMahdal: The `<description>` is never used.

AloisMahdal:
Also, whatever you meant to do, this is  **not** refactoring.

common.sh
```diff
@@ -160,9 +175,247 @@ exit_fixed()
...
+    # Add postupgrade script to /root/preupgrade/postupgrade.d directory
+    # File as first parameter has to exists in modules directory
+    # Usage:    add_postupgrade file
+
+    script_name=$1
+    postupgrade_name="$PREUPG_POSTUPGRADE_DIR/`basename $script_name`"
+    [ ! -f "$script_name" ] && log_error "$script_name does not exist." && exit_error
+    [ -f "$postupgrade_name" ] && log_warning "$script_name already exists in $PREUPG_POSTUPGRADE_DIR"
+    cp -af $script_name $postupgrade_name
+}
```
AloisMahdal: postupgrade.d was meant to be replaced by new hook system proposed as [rhbz#1334903][1] (and partially implemented in #108).

So unless I'm wrong, this function should not be needed.

  [1]: https://bugzilla.redhat.com/show_bug.cgi?id=1334903

common.sh
```diff
@@ -160,9 +175,247 @@ exit_fixed()
...
+    # Description of filename as second parameter which will be inserted into README file.
+    # Format in README file is:
+    # * <filename> - <description>
+    # Usage:    add_to_kickstart_readme <filename> <description>
+    filename=$1
+    description=$2
+    [ -z "$1" ] && log_error "Missing filename parameter."
+    [ -z "$2" ] && log_error "Missing description of filename to kickstart README."
+    echo " * $filename - $description" >> $KICKSTART_README
+}
```
AloisMahdal: IIUC, user will need two steps: adding file to KS (`add_kickstart_dir())`, and describing the file `(add_kickstart_readme())`.

Shouldn't one step be enough?

With a bit of shuffling, one variadic function can handle even more rich scenarios.  E.g. (naive code):

    keep_for_kickstart() {
        #
        # Store and keep file(s) for kickstart installation
        #
        # Usage:
        #
        #     keep_for_kickstart <description> <file>...
        #
        # Keep one or more <file>s for kickstart installation. <description> is mandatory and
        # will appear in README provided along with stored files.
        #
        local description=$1; shift
        local item
        local list_open=false
        {
            echo
            echo -n " * "
            for item in "$@"; do 
                # some checks...
                cp -ar $item $KS_STORAGE
                $list_open && echo -n ", "            
                echo -n "$(basename "$item")"; list_open=true
            done
            echo " - $description"
        } >> $KS_FILE
    }

Note that with above API, user can not just do both storing and documenting in one step, but also share description among multiple files.

common.sh
```diff
@@ -307,6 +308,77 @@ add_to_kickstart_readme()
...
+add_config_file()
+{
+    # Add config file to /root/preupgrade/{clean,dirty,etc} directory
+    # First paramater is configuration directory where we want to store a file.
+    #
+    # second parameter is a file which we store
+    # Usage:    add_config_file <clean,dirty> <full_path_file_name>
+
+    config_type=$1
+    filename=$1
```
jmazanek: line 331 - This appears to be a typo, you probably meant: filename=$2

AloisMahdal:
A more general advice (sorry for the essay; i really feel that this could save work on all sides). Try to avoid mentioning any implementation details (files, dirs...) in documentation, messages and/or function names.

A clean API does as little as possible, and exposes **only** terms that are relevant to the business logic.  Following this rule makes API both easier to understand but also easier to document and maintain.

For example:

    add_config_file()
    {
        # Add config file to /root/preupgrade/{clean,dirty,etc} directory
        # First paramater is configuration directory where we want to store a file.
        #
        # second parameter is a file which we store
        # Usage:    add_config_file <clean,dirty> <full_path_file_name>
        ....
 
This seems to make sense when you already know and care about the implementation details (eg. that files are copied around and kept in separate directories.), but the actual business logic has nothing to do with that. The core of business logic here is *assessment*; that is, module only wants to say "this file is dirty" or "this file is clean".  No more and no less.

Compare the above with:

    mark_file_as()
    {
        # Mark file as "clean" or "dirty".
        # First paramater is state, second parameter path to the file in question.
        #
        # Usage:    mark_file_as <clean,dirty> <full_path_file_name>
        #
        # Files marked this way will be stored for later review (dirty) or deployment (clean).
        ....
 
By leaving out directories:
 *  we don't have to describe them,
 *  we can change the paths without need to change docstring,
 *  the docstring and the API looks much more relevant --- it's now *obvious* why and how
    the API should be used.

AloisMahdal:
One last thing:  In Bash part, you're never using `local`; you're effectively introducing a bunch of globals for no reason.

common.sh
phracek: How do you think to get rid of them?
They has to exist in common.sh otherwise contents will not see them.

common.sh
phracek: I do not want to repeat test for existing file or command more then once.
See what you proposed:

    get_cached() {
        #
        # Get general system information from cache.
        #
        # Usage:  get_cached file <PATH>
        #         get_cached command <KEY>
        #         get_cached filelist <KEY>
        #         get_cached info <KEY>
        #
        # First sub-command is used to get cached version of file
        # that was originally on <PATH>.  Following paths are
        # currently cached:
        #
        #  *  /etc/passwd
        #  *  /etc/group
        #
        # Second sub-command `command` retrieves cached output of
        # a common system interrogation command identified by `<KEY>`,
        # which can be one of:
        #
        #  *  rpm_qa - output of `rpm -qa`
        #  *  rpm_Va - output of `rpm -Va`
        #
        # Third sub-command `filelist` gets cached file list named
        # <KEY>.  Format is one path per line and currently provided
        # lists are:
        #
        #  *  executable - (describe meaning)
        #  *  allmyfiles - (describe meaning)
        #
        # Last sub-command `info` provides access to miscellaneous
        # information cached by Preupgrade Assistant.  Valid keys
        # are:
        #
        # *  foo - (describe meaning and format)
        # *  bar - (describe meaning and format)
        #
        true
    }

common.sh
phracek: Yeah, it is a typo. It should be deleted if it does not exist.

common.sh
phracek: This has to be discussed with @pirat89 . We have an idea in case of duplicity of configuration files.

common.sh
phracek: We need to work in current directory. There are solution text etc. Therefore the function exists.

common.sh
phracek: right, also with add_to_kickstart_readme. I will fix it.

common.sh
phracek: Sure, but it was another Pull Request. And this Pull Request takes a pretty much time.

common.sh
phracek: Thanks

common.sh
phracek: And of course reason was from backwards compatibility point of view.

common.sh
phracek: store_to_kickstart is better but keep_for_kickstart. I would prefer store

phracek:
May be add_conf_file is really weird. But mark_file_as does not sound good as well.
What about store_config_file? and the rest is valid. I have already implemented in such way. First parameter is "clean" and second is full path.

common.sh
AloisMahdal: That does not really explain why any code would ever need to *leave* the directory (even using other than pushd/popd/), thus need this function.

common.sh
```diff
@@ -250,10 +250,10 @@ add_postupgrade()
     local postupgrade_name
 
     script_name=$1
-    postupgrade_name="$PREUPG_POSTUPGRADE_DIR/`basename $script_name`"
+    postupgrade_name="$PREUPG_POSTUPGRADE_DIR/$(basename $script_name)"
     [ ! -f "$script_name" ] && log_error "$script_name does not exist." && exit_error
     [ -f "$postupgrade_name" ] && log_warning "$script_name already exists in $PREUPG_POSTUPGRADE_DIR"
-    cp -af "$script_name" "$postupgrade_name"
+    cp -af ${script_name} ${$postupgrade_name}
```
AloisMahdal: The rightmost argument is apparently a typo.  But why are you removing quotes from the `"$script_name"`? The same for other places in this commit.

All in all, I believe this commit does not "fix" or improve the code, rather the opposite.

pirat89:
about @AloisMahdal  argument about docstring and bussiness logic:
That's true, that mentioned paths in functions like: "store file to ..." isn't the happiest text, because of hook system. There is explained that file will be stored to old directory, however there is already merged new hook system, which will store other postupgrade scripts into path /root/preupgrade/hook/....  so this will be not relevant in future I guess. So we should really just describe out expactations what function should do, like: ,,shedule processing of script during postupgrade phase".  In addition, when we say - it will be stored somewhere - it will mean, that changed path will be same like changed API. With just "bussiness" description, it will be still same from users point of view, fully compatible when use correctly the API.

common.sh
```diff
@@ -160,6 +163,243 @@ exit_fixed()
...
+    # Usage:    add_postupgrade file
+    local script_name
+    local postupgrade_name
+
+    script_name=$1
+    postupgrade_name="$PREUPG_POSTUPGRADE_DIR/$(basename $script_name)"
+    [ ! -f "$script_name" ] && log_error "$script_name does not exist." && exit_error
+    [ -f "$postupgrade_name" ] && log_warning "$script_name already exists in $PREUPG_POSTUPGRADE_DIR"
+    cp -af ${script_name} ${$postupgrade_name}
+}
```
pirat89: That should be changed. If I remember correctly, somewhere exists already RFE for that function with described behaviour. In addition, there is implemented hook system already, so not sure if hook is just prepared for next cycle and will be ignored yet or this should be completely rewritten. For me now:
 - just put all file(s) into the own directory (and create relevant directory if doesn't exist)
 - support multiple files in argument (support even directories: "cp -r")
 - check that file is executable (at least first - some postupgrade-scripts needs own data files which don't have to be executable

it is easier now for implementation and it will not case troubles with contents now. We should discuss it yet, to verify what we will able to modify in modules now.

common.sh
```diff
@@ -160,6 +163,243 @@ exit_fixed()
...
+    # Usage:    add_manual_postupgrade file
+
+    local script_name
+    local postupgrade_name
+    script_name=$1
+    postupgrade_name="$NOAUTO_POSTUPGRADE_D/$(basename $script_name)"
+    [ ! -f "$script_name" ] && log_error "$script_name does not exist." && exit_error
+    [ -f "$postupgrade_name" ] && log_warning "$script_name already exists in $NOAUTO_POSTUPGRADE_D"
+    cp -af ${script_name} ${postupgrade_name}
+}
```
pirat89: as comment above. not sure about usecase in hooks (probably this is outside of hooks), but in future probably could be created again own directory to make sure other scripts will not be replaced

common.sh
pirat89: in addition, in case that script already exists, function shouldn't replace it by another, but should "return 1" or exit with error too - think about case that some postupgrade script resolves kernel panic and this function it will remove it. this should not happen.

common.sh
```diff
@@ -160,6 +163,243 @@ exit_fixed()
...
+        *)
+            log_error "Unknown '$config_type' option in function add_config_file."
+            log_error "Allowed options are <clean,dirty,unchecked,deprecated>."
+            return 1
+            ;;
+    esac
+    mkdir -p $(dirname "$backup_file")
+    cp -a "$filename" "$backup_file"
+    return 0
+}
```
pirat89: function is wrong. it must check always all directories (not just one for one case, two for second case, ...).. I am able to copy one config file into more directories without any warning. Logic of this function will be little complicated, e.g.:
 - there is missing support for modified config files.
 - missing debug messages - for me will be really useful when I will be able to find a content which store a config file

And first two messages are switched.

phracek:
Personally, I don't see or even believe, we are able to provide API with this cycle.

phracek:
Please keep comments simple as soon as possible.

bocekm:
Can one of the admins verify this patch?

bocekm:
[test]

bocekm:
[test]

bocekm:
We have decided to not pursue finalizing this Pull Request any further.

PR #119, opened 2016-07-22
Shellfu style for Bash modules

### LINKED ISSUE #102: Use Shellfu coding style for Bash API
There aren't many coding styles for Bash, which (since I write a lot
id Bash) is the reason I had to create my own as part of Shellfu
project, so if I may humbly propose:

    http://file.vornet.cz/tmp/shellfu-style.html

Pros:

 *  Not big (~400-500 sparse lines), plus not all would be relevant
    to common.sh --- mainly the Docstrings section is relevant.

 *  Part of Shellfu project is a tool `shellfu-doc` which can help
    export docs out of the code that uses this style.

    (The tool is not finished --- it can't parse variable docs but
    it's a top priority to be done.)

 *  Really simple, does not get in the way.

 *  Quite similar to what's already used in common.sh.

Cons:

 *  Not 100% complete yet (although I don't expect it to grow much)

 *  Used only by me so far :) ... on multiple projects, though...

AloisMahdal:
Main goal is to resolve #102.

Also, "lost" documentation strings are ported from the abandoned PR #97 and some strings have been added by me; where I dared to guess.

To view and/or generate doc with `shellfu-doc` (missing features mentioned in the issue#102 are already implemented):

 1. install package *shellfu-devel* from [my Fedora Copr][1].

  [1]: https://copr.fedorainfracloud.org/coprs/amahdal/shellfu/

 2. `shellfu-doc ./common.sh` to view manpage directly

 3. `shellfu-doc -e pod ./common.sh` to export to POD (for use with eg. pod2man)

Also you can e.g.:

    shellfu-doc --help
    shellfu-doc --lsfun ./common.sh
    shellfu-doc --lsvar ./common.sh
    shellfu-doc -e markdown ./common.sh | Markdown > common.html

### DIFF
--- a/common.sh
+++ b/common.sh
@@ -1,3 +1,12 @@
+#!/bin/bash
+
+#
+# preupgrade-assistant module API -- Bash port
+#
+# These variables and functions will be available
+# for Bash preupgrade-assistant modules.
+#
+
 
 CACHE=/var/cache/preupgrade
 
--- a/common.sh
+++ b/common.sh
@@ -56,8 +56,7 @@ export LC_ALL=C
 #
 # Note that if env variable $COMPONENT is defined, it may be omitted from
 # parameters.
-log()
-{
+log() {
     SEVERITY=$1 ; shift
     if test -z "$COMPONENT"; then
         # only message was passed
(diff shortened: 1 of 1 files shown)


AloisMahdal:
Note that this is all just in Bash API;  the docstrings I have added were not added in Python. We'll need to do sync.

AloisMahdal:
Few more notes:

 *  The manpage formatting may look confusing for variables/functions that have missing docstrings (they look "grouped").  Future *shellfu-doc* version will address this by including placeholder text but in our case the strings will have to be filled in anyway.

 *  The format of docstrings should be easy to copy but in case you wanted to see exact specs, you can look on functions inside *shellfu-doc* ...  actually ... I believe `shellfu-doc /usr/local/bin/shellfu-doc` would work as well!

phracek:
ok, thanks. I will port it into Python part before a release.

PR #121, opened 2016-07-28
Improve component name placeholder

AloisMahdal:
Bash API uses '[unknown]' so Python API should do the same.

### DIFF
--- a/common.sh
+++ b/common.sh
@@ -223,7 +223,7 @@ log() {
     if test -z "$COMPONENT"; then
         # only message was passed
         if test "$#" -eq 1; then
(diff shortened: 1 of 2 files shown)


AloisMahdal:
Another improvement (the second commit):

Currently, if component is unknown, the resulting log line may be
confusing, e.g.:

    INFO [unknown]: foo bar baz

In this form it's not clear *what* is unknown; the line feels ambiguous.
The new form:

    INFO [unknown-component]: foo bar baz

makes it clear that it's the component name that is missing, without
need to study details of logging format.

PR #122, opened 2016-07-29
Improve logging format

AloisMahdal:
Note: this is addressed by [bug 1361378][1] in downstream.  Following text is taken from there.

  [1]: https://bugzilla.redhat.com/show_bug.cgi?id=1361378

Description of problem
======================

Currently there are two issues with logging from modules:

 *  Risks use prefix "INPLACERISK:" that is not suitable for migration
    scenarios.

 *  Other log messages use format that is hard to filter out.

Together with bug 1309491 (openscap merges stderr to stdout), this makes
it hard to catch situations when scripts leak some info outside intended.

For example, "leaking" grep as in bug 1356811, sntax errors or errors
thrown by external utilities called by modules are all problems
that could be easily checked by tests *if* the formatting was more
filtering-friendly.

Proposal
========

For logging risks, use format:

    preupg.risk.MEDIUM: some risky risk

For other properly logged messages, use:

    preupg.log.INFO: some_component: I just want to let you know

This way, all properly logged messages and risks can be easily separated
from leaks or errors thrown for various reasons.

### DIFF
--- a/common.sh
+++ b/common.sh
@@ -233,7 +233,7 @@ log() {
         fi
     fi
 
-    echo "$SEVERITY $COMPONENT: $1" >&2
+    echo "preupg.log.$SEVERITY: $COMPONENT: $1" >&2
 }
 
 log_debug() {
--- a/common.sh
+++ b/common.sh
@@ -297,7 +297,7 @@ log_risk() {
     #
     # log risk level to stderr
     #
-    echo "INPLACERISK: $1: $2" >&2
+    echo "preupg.risk.$1: $2" >&2
 }
 
 log_none_risk() {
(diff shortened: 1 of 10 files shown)


AloisMahdal:
Note: I replaced the formatting also in various auxiliary scripts (preupg_ui.py, tests...) but **I expect to have missed something**!

phracek:
I would like to remove 'preupg.' at the beginning of message. Rest of is fine from my point of view.

I don't think, 'preupg.' is really needed.

phracek:
@AloisMahdal What do you think about it? Are you able to update Pull Request?

AloisMahdal:
We had a chat about this w/ @phracek, but FTR:  The `preupg.` prefix is there to safely identify messages generated by *preupg* (or its modules).

Removing it would remove fix of the second problem I mentioned in the OP.

PR #126, opened 2016-08-12
Hide generic logging message functions

AloisMahdal:
In Python, it's convention to name private stuff starting with underscore.

In Bash, shellfu-doc respects this convention as well, so this will hide
these functions from a generated manpage or listing.


pirat89:
Agree. This seems good to me.

PR #127, opened 2016-08-12
Don't pollute Bash environment with variables

AloisMahdal:
This is a functional change; now we re-set these global variables every
time someone uses any of those functions.  Of course, nobody should, but
in theory, someone could rely on this (due to strange bug or naming
conflict).

### DIFF
--- a/common.sh
+++ b/common.sh
@@ -219,7 +219,7 @@ log() {
     # Note that if env variable $COMPONENT is defined, it may be omitted from
     # parameters.
     #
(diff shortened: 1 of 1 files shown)


pirat89:
Agree. Good catch! Mainly for $pkg variable which is usually used too. Just there is missing one another variable "line" which need to be local too.

AloisMahdal:
Thanks, @pirat89.  Updated the commit.

PR #144, opened 2016-09-16
Fix VALUE_RPM_QA line parsing

AloisMahdal:
Previous commit (9bc7cc2 - bash quoting) caused a long-standing problem to propagate.  Previous code:

    pkg=$(echo $line | cut -d ' ' -f1)

worked only "accidentally".  The code apparently assumed that field delimiter in the line (from VALUE_RPM_QA) is space (`' '`).

This is not true; the lines are delimited with tab, but since the `$line` referenc was was not quoted, this resulted in multiple arguments passed to `echo`, which in turn joins all arguments with single *space*. In other words, the tabs got effectively converted to spaces.

Commit 9bc7cc2 added quoting, but this meant that the "conversion" did not happen anymore so the package ended up being something like `foo\tRed` (since the line was `foo\tRed Hat...`).  Needless to say, this completely broke the `get_dist_native_list()` function.

This commit fixes the situation by ensuring that only the first field (before any kind of space) makes it into the `$pkg` variable.


AloisMahdal:
Tested by replacing common.sh in the last downstream build and running downstream 'nativity' test.

PR #152, opened 2016-09-21
Deploy the hook files to correct path (just as Bash version)

AloisMahdal:


phracek:
Finally, I have recognized the problem with xccdf_ definitions. It seem, like MODULE_PATH has to contain xccdf_ prefix too and code will not be so complicated.

AloisMahdal:
Or even better, why do you need the `xccdf_` prefix in the first place?

PR #156, opened 2016-10-10
Add default value for check script

phracek:

This Pull Request adds a default value in case user did not specify it.


phracek:
If user does not enter anything, then SH type is used.
If user specify anything else then sh, py then message like "sh or py should by specify"

The rest is ok.

phracek:
@pirat89 @bocekm Please review it. Relevant bugzilla  https://bugzilla.redhat.com/show_bug.cgi?id=1332792

bocekm:
Peter, you changed options to be shown to the user to [SH/py] (settings.py:42), i.e. you changed SH to uppercase. However the options in ui_helper.py:141 are still both lowercase ['sh', 'py']. The same with the warning on ui_helper.py:147, which says "Select either 'sh or 'py'." That means if the uses types uppercase "SH" as they were asked, the script will warn them to type lowercase "sh" instead.

PR #162, opened 2016-11-01
Implementation postmigrate.d directory

phracek:

This Pull Request will execute user defined scripts mentioned in "postinstall.d" directory on migrated system.

User/admin is inform in case "postinstall.d" is empty.

~~~
# preupg-kickstart-generator 
You did not specify any postinstall scripts. Include them to /root/preupgrade/postinstall.d 

Do you want to continue with kickstart generation without postinstall scripts? (Y/n)
~~~

### DIFF
--- a/kickstart/finish.sh
+++ b/kickstart/finish.sh
@@ -30,7 +30,7 @@ do
     fi
 done
 cd ${PWD}
-POST_INSTALL_SCRIPTS="postinstall.d"
+POST_INSTALL_SCRIPTS="postmigrate.d"
 cd $POST_INSTALL_SCRIPTS
 for file in $(find . -type f -executable)
 do
--- a/kickstart/finish.sh
+++ b/kickstart/finish.sh
@@ -30,8 +30,8 @@ do
     fi
 done
 cd ${PWD}
-KS_SCRIPTS="kickstart/scripts"
-cd $KS_SCRIPTS
+POST_INSTALL_SCRIPTS="postinstall.d"
+cd $POST_INSTALL_SCRIPTS
 for file in $(find . -type f -executable)
 do
     echo "Running script ${file} ..." >> ${PREUPGRADE_LOG}
--- a/preup/kickstart/application.py
+++ b/preup/kickstart/application.py
@@ -212,7 +212,7 @@ class KickstartGenerator(object):
                 kickstart_data[index] = "#" + row
         FileHelper.write_to_file(self.kick_start_name, 'wb', kickstart_data)
 
-    def check_postimigrate_dir(self):
+    def check_postmigrate_dir(self):
         if not FileHelper.get_list_executable_files_in_dir(os.path.join(settings.result_dir,
                                                                         settings.postmigrate_dir)):
             if not self.conf.assumeyes:
@@ -232,7 +232,7 @@ class KickstartGenerator(object):
         if not self.collect_data():
             log_message("Important data are missing for the Kickstart generation.", level=logging.ERROR)
             return None
-        if not self.check_postimigrate_dir():
+        if not self.check_postmigrate_dir():
             return None
         self.ks.handler.packages.excludedList = []
         self.plugin_classes = self.load_plugins(os.path.dirname(__file__))
--- a/preup/kickstart/application.py
+++ b/preup/kickstart/application.py
@@ -196,13 +196,14 @@ class KickstartGenerator(object):
         list_issues = [' --', 'group', 'user ', 'repo', 'url', 'rootpw']
         kickstart_data = []
         try:
-            kickstart_data = FileHelper.get_file_content(os.path.join(settings.KS_DIR, self.kick_start_name),
+            kickstart_data = FileHelper.get_file_content(os.path.join(settings.KS_DIR,
+                                                                      self.kick_start_name),
                                                          'rb',
                                                          method=True,
                                                          decode_flag=False)
         except IOError:
-            log_message("The %s file is missing. The partitioning layout might not be complete." % self.kick_start_name,
-                        level=logging.WARNING)
+            log_message("The %s file is missing. The partitioning layout "
+                        "might not be complete." % self.kick_start_name, level=logging.WARNING)
             return None
         for index, row in enumerate(kickstart_data):
             tag = [com for com in list_issues if row.startswith(com)]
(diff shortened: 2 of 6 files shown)


phracek:
Of course scripts needs to have executable rights.
Relevant issue #162

bocekm:
We should discuss where to document features like this. I think man page would be the right place.
This pull request is the best place where to introduce the updated documentation. If we don't document new feature/option/behavior right with its implementation, then there is a significant risk of not doing it later.

preup/kickstart/application.py
```diff
@@ -210,10 +210,28 @@ def comment_kickstart_issues(self):
                 kickstart_data[index] = "#" + row
         FileHelper.write_to_file(self.kick_start_name, 'wb', kickstart_data)
 
+    def _get_all_postinstall_scripts(self):
```
bocekm: 1. This method can be static as it does not handle any class data. It can be also part of utilities if implemented as the second point below suggests.
2. Why is this method named *_get* something, when the only thing you are using it for is the information whether there is/there is not at least one script in that particular folder. It seems to me boolean output would fit here. And the name like: *_contains_executable_files(dir)*

preup/kickstart/application.py
```diff
@@ -210,10 +210,28 @@ def comment_kickstart_issues(self):
...
+                file_name = os.path.join(path, f)
+                if os.path.exists(file_name) and FileHelper.check_file(file_name, u"x") is True:
+                    found_scripts.append(file_name)
+        return found_scripts
+
     def generate(self):
         if not self.collect_data():
             log_message("Important data are missing for the Kickstart generation.", level=logging.ERROR)
             return None
+        if not self._get_all_postinstall_scripts():
```
bocekm: The implementation of this whole if condition is quite low level (level of abstraction). However the method generate is very high level. We should not mix these different levels in one function/method. Please move this condition into a separate method. Then just one line like the following would be here: "

preup/kickstart/application.py
```diff
@@ -210,10 +210,28 @@ def comment_kickstart_issues(self):
...
+            accept = ['y', 'yes']
+            log_message("You did not specify any postinstall scripts. "
+                        "Include them to %s " % os.path.join(settings.result_dir, settings.postinstall_dir))
```
bocekm: *"You did not specify any postinstall scripts. Include them to .."* sounds like the user should have done that. However this step is optional.
I would rephrase it to *The %s folder is empty - scripts to be executed after the migration should be placed here. ... Do you want to continue with the kickstart generation without any postmigration scripts?"

preup/kickstart/application.py
```diff
@@ -210,10 +210,28 @@ def comment_kickstart_issues(self):
...
+            choice = MessageHelper.get_message(message="Do you want to continue with kickstart generation "
```
bocekm: Please stick with the maximum of 80 chars per line PEP rule.

preup/kickstart/application.py
```diff
@@ -210,10 +210,28 @@ def comment_kickstart_issues(self):
...
+                                               "without postinstall scripts?",
+                                               prompt="(Y/n)")
+            if choice.lower() not in accept:
+                return None
```
bocekm: The *generate* method is called from the main method and when None is returned here, the main method continues in executing. Is it correct/expected behavior? Shouldn't the main return as soon as it receives None from here, with some log message ideally?

bocekm:
Also please address the failed tests. Thanks.

kickstart/finish.sh
```diff
@@ -30,8 +30,8 @@ do
     fi
 done
 cd ${PWD}
-KS_SCRIPTS="kickstart/scripts"
-cd $KS_SCRIPTS
+POST_INSTALL_SCRIPTS="postinstall.d"
```
phracek: Make sense, I will change it.

phracek:
Output now looks like:

~~~
[root@rhel68 ~]# preupg-kickstart-generator 
The '/root/preupgrade/postmigrate.d' folder is empty - scripts to be executed after the migration should be placed here.

Do you want to continue with kickstart generation without any postmigration scripts? (Y/n)

The Preupgrade Assistant generates a kickstart file in '/root/preupgrade/kickstart/preupgrade.ks'.
The Kickstart file contains:
- users with UID/GID which you should create on Red Hat Enterprise Linux 7 system.
- the partitioning layout which was used on this system
- the package set which was installed on this system.
- the firewall rules which were enabled on this system.
The Kickstart file is pre-generated from this system and is not to be used directly for 
the installation of Red Hat Enterprise Linux 7.
The Kickstart file needs to be modified by the administator.
These directories exist on the migration system:
- /root/preupgrade/preupg_results-161101125919/cleanconf - configuration files which can be used on the migrated system.
- /root/preupgrade/preupg_results-161101125919/dirtyconf - configuration files which need to be clarified by the administrator.

[root@rhel68 ~]# 
~~~

And with --force option
~~~
[root@rhel68 ~]# preupg-kickstart-generator --force
The Preupgrade Assistant generates a kickstart file in '/root/preupgrade/kickstart/preupgrade.ks'.
The Kickstart file contains:
- users with UID/GID which you should create on Red Hat Enterprise Linux 7 system.
- the partitioning layout which was used on this system
- the package set which was installed on this system.
- the firewall rules which were enabled on this system.
The Kickstart file is pre-generated from this system and is not to be used directly for 
the installation of Red Hat Enterprise Linux 7.
The Kickstart file needs to be modified by the administator.
These directories exist on the migration system:
- /root/preupgrade/preupg_results-161101125919/cleanconf - configuration files which can be used on the migrated system.
- /root/preupgrade/preupg_results-161101125919/dirtyconf - configuration files which need to be clarified by the administrator.

[root@rhel68 ~]# 
~~~

preup/kickstart/application.py
phracek: This is catched by MessageHelper. You have to answer either Y or N. Nothing else is accepted.

phracek:
Thanks for the review. This is the way, how the project can be useful for all folks. I am glad for reviews.

preup/kickstart/application.py
bocekm: I mean something a bit different. When the user answers N, I would expect that the kickstart generation will be stopped (after the question "Do you want to continue with kickstart generation..?"). But instead the generate function returns None and the script continues to run.

preup/kickstart/cli.py
```diff
@@ -12,12 +12,20 @@ def __init__(self, args=None):
...
-        #self.add_args()
+        self.add_args()
         if args:
             self.opts, self.args = self.parser.parse_args(args=args)
         else:
             self.opts, self.args = self.parser.parse_args()
 
+    def add_args(self):
+        self.parser.add_option(
+            "--force",
```
bocekm: I don't think that force is the best name here. I would use the same as at least yum/dnf has, i.e. -y option, which makes all the questions to be answered yes automatically.

preup/kickstart/application.py
```diff
@@ -210,28 +210,22 @@ def comment_kickstart_issues(self):
...
-                                               prompt="(Y/n)")
-            if choice.lower() not in accept:
-                return None
+        if not self.conf.force:
+            if not FileHelper.get_list_executable_files_in_dir(os.path.join(settings.result_dir,
+                                                                            settings.kickstart_dir)):
+                accept = ['y', 'yes']
+                log_message("The '%s' folder is empty - scripts to be executed "
+                            "after the migration should be placed here." % os.path.join(settings.result_dir,
+                                                                                        settings.postmigrate_dir))
```
bocekm: This line is totally out of 80 chars per line. See how it looks like in my editor [1] - totally unreadable.
[1] https://postimg.org/image/jclxer9d3/
Here's one of the ways how it can be written with the 80 chars limit:
```python
                postmigrate_path = os.path.join(settings.result_dir,
                                                settings.postmigrate_dir)
                log_message("The '%s' folder is empty - scripts to be"
                            " executed after the migration should be placed"
                            " here." % `postmigrate_path)
```

preup/utils.py
```diff
@@ -280,6 +280,22 @@ def get_script_type(file_name):
...
+        The function returns list of executable files in directory.
+
+        :param dir_name: Dir name where executable files are searched
+        :return: List of files or empty list
+        """
+        found_scripts = []
+        for (root_path, dirs, files) in os.walk(dir_name):
+            for f in files:
+                file_name = os.path.join(root_path, f)
+                if os.path.exists(file_name) and FileHelper.check_file(file_name, u"x") is True:
```
phracek: Without unicode it does not work. I would prefer to use u"x" which indicates always unicode as for Python 2 as for Python 3.

bocekm: Oh, so the problem is on [this line of check_file](https://github.com/upgrades-migrations/preupgrade-assistant/blob/kickstart_postinstall_dir/preup/utils.py#L94) function. It requires Unicode string, but that makes no sense. Please update the condition there to:
`if isinstance(mode, six.string_types):`

preup/kickstart/application.py
```diff
@@ -196,36 +196,43 @@ def comment_kickstart_issues(self):
...
         FileHelper.write_to_file(self.kick_start_name, 'wb', kickstart_data)
 
-    def generate(self):
-        if not self.collect_data():
-            log_message("Important data are missing for the Kickstart generation.", level=logging.ERROR)
-            return None
-        if not self.conf.force:
-            if not FileHelper.get_list_executable_files_in_dir(os.path.join(settings.result_dir,
-                                                                            settings.kickstart_dir)):
+    def check_postimigrate_dir(self):
```
bocekm: Probably a typo in the function name? :)

bocekm:
I can see my comment from above that hasn't been done in this PR and I regret I didn't recalled it and approved the PR:
"We should discuss where to document features like this. I think man page would be the right place.
This pull request is the best place where to introduce the updated documentation. If we don't document new feature/option/behavior right with its implementation, then there is a significant risk of not doing it later."

PR #172, opened 2016-11-23
Text correction of TXT file of template01

jhornice:


templates/01_simple_informational_module/solution.txt
```diff
@@ -1,17 +1,17 @@
...
-this one, instructions to resolve troubles, etc.
+This is a solution text file that contains detailed information that will
+be a part of the report file. You can put here an informational message, like
+this one, or some instructions to resolve issues.
 
-The file can be empty and text can be generated by the check script itself.
+The file can stay empty and the text can be generated by the check script itself.
 Usually, you put here information that is static and you don't want
-to generate it by the script each run again and again.
+to generate it by the script every time you run it.
```
AloisMahdal: The formulation seems to imply that reader of this text ("you") is going to be the one that runs the module, which is usually not the case.

PR #180, opened 2016-11-23
Provide examples instead of templates, how to write modules

pirat89:
Templates for modules doesn't make sense and these were really odd. Instead of templates,
examples will be provided which will show, how to write modules and use Preupgrade Assistant API
in simple way. This set now contains 4 examples which show simple way how to write basic modules.

Another could be added in future, this is just replacement of previous templates. It should be well
understandable and everyone will be able to know how to write another similar simple modules
based on these. I guess, in future some examples could show, how to write modules in Python. But it should not copy exactly current examples, but provide something new, little more advanced maybe.


doc/module_examples/02_inspection_needed/check.sh
```diff
@@ -0,0 +1,50 @@
...
+. /usr/share/preupgrade/common.sh
+
+#END GENERATED SECTION
+
+#
+# The whole section above is processed and modified by preupg-xccdf-compose
+# script, according to content of INI file (in this case content.ini).
+# And in addition the LICENSE used by PreupgradeAssistant is inserted.
+#
+
```
jmazanek: It is a nitpick, but part of the above section is also a shebang line. The wording as it is indicates that it is also provided by preupg-xccdf-compose. As far as I know it is not true as shebang line should be included right when writing module code (for good reasons). Please consider rewording to reflect that.

doc/module_examples/03_action_required/check.sh
```diff
@@ -0,0 +1,47 @@
...
+# is inserted.
+#
+
+##
+# Briefly:
+#   In case a manual inspection, check or action BEFORE the upgrade is needed,
+#   do the following things:
+#     1) provide a text in $SOLUTION_FILE: a description of the problem and
+#        remediation instructions
+#     2) use "log_high_risk" to provide a short message that a problem was
```
bocekm: Does it really need to be exactly "log_high_risk" if I need to request user to do something before the upgrade? I mean it in a sense that it can't be "log_medium_risk" for example.

- If **so**, reword this bullet to something like:
*2) use "log_[severity]_risk" to provide a short message that a problem was found. Change [severity] to one of the levels of severity available through API, for example "log_medium_risk"*

- If **not**, you can prevent others asking this question :) For example by adding something like *"Preupgrade Assistant requires the risk to be set to high in this scenario"*

doc/module_examples/04_fix_issue/check.sh
```diff
@@ -0,0 +1,168 @@
...
+
+. /usr/share/preupgrade/common.sh
+
+#END GENERATED SECTION
+
+#
+# The whole section above is processed and modified by the preupg-xccdf-compose
+# script, according to the contents of the INI file (in this case it is the
+# content.ini file). In addition, the LICENSE used by the Preupgrade Assistant
+# is inserted.
```
bocekm: This note is duplicate. Because this is a tutorial and the user goes through all the parts, starting with 01, I think it can be specified just in the 01 part.

doc/module_examples/04_fix_issue/check.sh
```diff
@@ -0,0 +1,168 @@
...
+    echo
+  } >> "$SOLUTION_FILE"
+  set_result $RESULT_FAIL
+fi
+
+# case 3)
+# ok, let's pretend that when a naughty-foo-cottage subpackage is installed,
+# we will want to install a naughty-foo-house on the new system
+if is_pkg_installed "naughty-foo-house"; then
+  log_info "The 'naughty-foo-cottage' package has been split on the new system. A 'naughty-foo-house' package will be installed."
```
bocekm: Split is not the best word here IMO. First time I saw that I was like "What that is suppose to mean? Is it some kind of jargon I don't know about?" :) Better would be *".. has been separated into two different packages .."*. Or use *divided* instead of *separated*, if you wish.

doc/module_examples/02_inspection_needed/check.sh
```diff
@@ -0,0 +1,54 @@
...
+#
+# The whole section above is processed and modified by the preupg-xccdf-compose
+# script, according to the contents of the INI file (in this case it is the
+# content.ini file). In addition, the LICENSE used by the Preupgrade Assistant
+# is inserted.
+#
+
+##
+# Briefly:
```
pirat89: I disagree. IMO it makes whole module much clear just from this piece of text. It whould be ideally inside each module >= 02. The reader will be able to skip it if it is not interested about this text and got down to the code.

doc/module_examples/03_action_required/check.sh
pirat89: Yes - it has to be log_high_risk - that's how PA resolves it. That's why I think it is good to write there - because it is not intuitive.

doc/module_examples/04_fix_issue/check.sh
AloisMahdal: "split" seems OK to me as it fits the "1 to N" operation.  Maybe it is a bit weird jargon, but seems familiar to me (just 3 years in Red Hat, though...)

That said, it's still not the most fortunate formulation. What about:

    log_info "Functionality of the 'naughty-foo-cottage' package has been split into 'naughty-foo-cottage' and 'naughty-foo-house'.  The 'naughty-foo-house' package will be installed to preserve all capabilities.."

(Frankly, I don't really like the "naughty" theme here. What's wrong with 'foo' and 'bar'?)

pirat89: @AloisMahdal about "naughty" - it is more "colourful" now, isn't it? :-D But seriously. I want there some difference against just foo & bar. Some name which says exactly: heyy! This is the bad guy! Foo & bar is not enough for me. You can discus it with jhornice. If something else could be used. I am not against rename. Just want something like this, what I described above.

doc/module_examples/02_inspection_needed/check.sh
bocekm: Ok

doc/module_examples/03_action_required/check.sh
bocekm: Not intuitive indeed. I still do not get why it can't be medium risk :( We need to focus on this when updating API.

Issue #143, opened 2016-09-15
preupg binary should provide a computer readable format


phracek:
After an assessment preupg binary creates by oscap XML and HTML files.
XML file is a pretty complicated because of parser.

It would be nice to add a functionality which will create a JSON file.
The JSON file will contain all assessments including results and risk which can be easily
parsed and human readable then XML file.

phracek:
@MajoDurco Can you please look on it? I would prefer JSON format.

AloisMahdal:
I don't quite understand this issue.

As you mentioned, preupg binary already provides a pretty computer readable format.  It parses well in almost anything since RHEL5 at least (in our downstream test library we are havily relying on it.)

And for humans, we already have HTML report, right?

So do you have a real use case for this?

phracek:
HTML is not suitable for comparison between two reports. XML provides a lot of garbage.
Translating XML file to JSON with correct information like

~~~
results: {'module_name': 'result': 'RESULT',
                                        'logs': [list of logs]}
~~~
Is easy to compare two reports. No garbage and these information we already have in P-A.
JSON file -> dictionary -> easy to compare in Python of Perl (for jmazanek).

First of all we have to provide and JSON file and then we will implement issue https://github.com/upgrades-migrations/preupgrade-assistant/issues/158

AloisMahdal:
Well it makes sense but still:

 *  First, we should avoid situation when there are two data files.

    If people are going to interpret results an automated way, there should be only one
    obvious authoritative source.  That source has been results.xml. If you add results.json
    next to it, chances are that you will end up with some people relying on XML and some
    on JSON.

 *  Next, don't forget that everything has to work on RHEL5.

 *  If the use case is comparison by humans, why JSON?  Why not just a simple text file?

     Again, by adding a file in a format which main purpose is serialization, you are sending
     out signal: "here's a data file you can parse and report".

Also, it seems to me that preupg is already doing too much.  As I have pointed out long time ago (one of my firs bug reports), the CLI is already confusing exactly because of this.

Ideally all that preupg does is generate assessment report (in a simple but rigorous and well-defined format) and that should be it.  You can have plethora of other tools that help *further* work with the report (analyze, filter, query, you name it...) which is where your RFE would fit much better, without bloating preupg even more.

PR #182, opened 2016-11-25
Add conversion xml->json

MajoDurco:


preupg/xml_to_json.py
```diff
@@ -0,0 +1,416 @@
...
+
+
+class MissingElement(Exception):
+    """
+    Element in xml file does not exist
+    """
+    pass
+
+
+class EmptyText(Exception):
```
bocekm: Name the exceptions so they end with either _Exception_ or _Error_, e.g. EmptyTextException.

bocekm:
As @AloisMahdal mentioned in the [Issue#143](https://github.com/upgrades-migrations/preupgrade-assistant/issues/143) and [bz#1398488](https://bugzilla.redhat.com/show_bug.cgi?id=1398488), we shouldn't be overloading preupg binary with that many tasks.
As we talked about it, your functionality proposed here should be executed independently on the Preupgrade Assistant assessment itself. Therefore, please move the code from _preupg/application.py_ to a new script within _tools_ directory. The implementation details are up to you, however here's my proposal:
- required input argument - path to the _result.xml_,
- optional input argument - output directory or path for the json file,
- name of the script be something like _preupg-xml-to-json_

bocekm:
Also please write unit tests for this new python module. Name it _tests/test_xml_to_json.py_.
In the past there hasn't been a big push to do that, but the team agreed it's a good practice to write unit tests with every new functionality.
It would be great if you include the tests to this pull request. However if you would rather submit them as a separate pull request, then write a new Issue for yourself (stating that tests for this functionality are needed) and work it as a next thing. Thanks.

bocekm:
If I remember well from a discussion with @phracek, it was QE team and @jmazanek who were supposed to use the JSON tooling.
@AloisMahdal says that he and the QE team don't need JSON output at all.
@jmazanek, do you still expect to use the JSON output for your multimachine applications? If not, I would even hesitate to integrate this into master branch, because it would basically become dead code.

jmazanek:
@bocekm, please keep JSON format at least as an option. It is easier to parse and handle than XML files. I will surely have a use case for it.

bocekm:
[test]

AloisMahdal:
@jmazanek Frankly, I don't understand what you or Petr mean by "JSON is easier to parse than XML".  How is that a problem? Of course, XML is much more complex but in either case you you're not parsing those formats yourself, are you?

Also you're saying "I will have use case for it" --- then I assume you are OK if we postpone adding of the feature until you do have the case?

I don't want to be a drag on this, but I believe we should prevent feature creep (especially when there are some architectural changes in contemplation) and this totally sounds to me as case of YAGNI.

pirat89:
@AloisMahdal frankly, even with libraries, work with json is much simpler and more intuitive then XML. However I personally don't need parse our result XML, so I don't need it. However I still think, that if someone want to parse it, it will much rather work with JSON then XML.

bocekm:
I kind of agree with @AloisMahdal, it looks like YAGNI. It would make sense to make it available once we have a real use case, not just "JSON is easier to parse, let's output it alongside XML and hope we'll use it someday."
BTW, I wrote [preupg-diff](https://github.com/upgrades-migrations/preupgrade-assistant/blob/preupg-diff/tools/preupg-diff) that parses the result.xml and it was pretty hassle-free and quick. If you take a look at the code you'll see that the parsing is really simple.
Anyway, I'll review the recent changes @MajoDurco has made and we can talk about it on a meeting, whether to leave it in a non-master branch till it's needed or merge it.

preupg/xml_to_json.py
```diff
@@ -0,0 +1,404 @@
...
+        By default converts:
+            from <TestResult>:
+                end-time(attribute), start-time(attribute)
+                inside:
+                    target-address(text)
+            from <rule-result>:
+                inside:
+                    result(text), set-value(idref), check-import(text)
+    '''
+    def __init__(self, file_name, additional_rule_elements=None, verbose=False):
```
bocekm: Please make use of the verbose parameter. It would be great if you could add a --verbose command line option for the user.

preupg/xml_to_json.py
```diff
@@ -0,0 +1,404 @@
...
+                        if self.verbose:
+                            log_message(err_mesg, level=logging.ERROR)
+            # {check: system} -> attribute system
+            else:
+                try:
+                    self.checkForValidAttribute(elem, attribute)
+                except AttrNotFoundException as err_mesg:
+                    element.remove(elem)
+                    if self.verbose:
+                        log_message(err_mesg, level=logging.ERROR)
```
bocekm: I'm getting the following traceback. It's because the function _log_message_ expects a string but you're passing an Exception object. You can fix it by:

_log_message(err_mesg.**message**, level=logging.ERROR)_
```
Traceback (most recent call last):
  File "./preupg-xml-to-json", line 30, in <module>
    main()
  File "./preupg-xml-to-json", line 27, in main
    json_file.write(convertor.convert())
  File "/home/mbocek/repos/rhel_pkgs/preupgrade-assistant-majo/preupg/xml_to_json.py", line 68, in convert
    self.processResults()
  File "/home/mbocek/repos/rhel_pkgs/preupgrade-assistant-majo/preupg/xml_to_json.py", line 148, in processResults
    rule_result_dict[element_idref] = self.parseDictOfElements(rule, self.rule_result_elements)
  File "/home/mbocek/repos/rhel_pkgs/preupgrade-assistant-majo/preupg/xml_to_json.py", line 172, in parseDictOfElements
    self.checkFoundElement(element, attribute)
  File "/home/mbocek/repos/rhel_pkgs/preupgrade-assistant-majo/preupg/xml_to_json.py", line 207, in checkFoundElement
    log_message(err_mesg, level=logging.ERROR)
  File "/home/mbocek/repos/rhel_pkgs/preupgrade-assistant-majo/preupg/logger.py", line 64, in log_message
    sys.stdout.write(message.encode(settings.defenc))
AttributeError: 'AttrNotFoundException' object has no attribute 'encode'
```

preupg/xml_to_json.py
```diff
@@ -0,0 +1,404 @@
...
+    def checkForValidAttribute(self, element, attribute):
+        '''
+        Check if text or attribute is empty or missing
+        if so make report about it and don't bother with it anymore
+        '''
+        if attribute == "text":
+            text = self.getDataFromElement(element, attribute)
+            if text is None or text.strip() == '':
+                raise AttrNotFoundException("Text in \"{0}\" element is empty doesn't exist"
+                                   .format(self.removeNamespace(element)))
```
bocekm: When I fix the verbose logging I'm getting huge number of exactly this sentence:
`Text in "check-import" element is empty doesn't exist`
It does not say much. I suggest:
```
raise AttrNotFoundException("Text in \"{0}\" element on line {1} is empty".format(self.removeNamespace(element), element.sourceline))
```
but for that you need to substitute
~~import xml.etree.ElementTree as ET~~
by 
`from lxml import etree as ET`

lxml is just improved ElementTree library (and it's available on both vanilla RHEL5 and RHEL6).

tools/preupg-xml-to-json
```diff
@@ -0,0 +1,30 @@
...
+import os
+import argparse
+
+from preupg import xml_to_json
+
+
+def main():
+    parser = argparse.ArgumentParser(prog='preupg-xml-to-json',
+                                     description='''
+                                     PreupgradeAssistant tool
```
bocekm: ~~PreupgradeAssistant~~ -> Preupgrade Assistant

bocekm: ~~PreupgradeAssistant~~ -> Preupgrade Assistant

jmazanek:
@MajoDurco I understand, that work on your preupg-xml-to-json tool is on hold currently, as it's implementation into the preupgrade-assistant project code is currently postponed as a (possible) feature in the future. But I find it really useful for my lab project. Is it currently usable as a standalone script, or could you make it so with a reasonable amount of work? 
For the record, this is not a push for its  inclusion into the main project, but should that ever return on the table, I am able to provide set of use cases and/or test scenarios for it

bocekm:
@MajoDurco, because @jmazanek has a use case for your tool, please take a look at it again, and once ready, we can push it to master. Thanks.

Issue #103, opened 2016-05-23
Utilize function(s) in API rather than bunch of globals


AloisMahdal:
The Problem
===========

Access to facilities like cached data or KS README appends is currently provided in rather inappropriate ways; read: by setting a bunch of variables to paths to rather nebulous files.

As a result, there are too many variables that are hard to even document properly.

Examples
--------

For example,

    # Full path to log file with all local files
    VALUE_ALLMYFILES=$PREUPGRADE_CACHE/allmyfiles.log

raises question how one could reliably use that:  "log" normally means time-stamped record of events; it's really misleading and unclear on how to parse that.  There are more: `.../chkconfig.log`, `.../passwd.log`, `.../postupgrade.d`.

Another one:

    # README file which contains description about all files in kickstart directory
    KICKSTART_README=$KICKSTART_DIR/README

...so what do I (module devel) do with that file?  As @pirat89 explained, this is so that path to KS README (to which my module is supposed to append to) is abstracted away in case it will change in future.

But what prevents module devel from eg. overwriting the file?

The Proposal
============

A better approach would be to provide set of functions.

What functions can do that variables cannot?  Functions enable:

 *  paths abstraction--just as well as variables do now,

 *  more expressive syntax (eg. `ks_add "my_file" "My description"`),

 *  read/write control  (eg. prevent clobbering, force read-only...),

 *  control over data presence (eg. append to KS README but also to HTML or some machine readable format),

 *  control over formatting (eg. when appending to KS README),

 *  error checking and/or debugging,

 *  less tokens--several related variables could be replaced by one function which means *less* stuff to document/explain/remember,

 *  more convenient usage (eg. in Python, calling a function is way more convenient than opening+parsing a file whose name is in `os.env`),

 *  even more convenient usage, (eg. in Python, lists can be lists).

phracek:
Here is first Pull Request #110 . Please do comments on it.

PR #186, opened 2016-12-02
Add guide for API consistency

AloisMahdal:
In order to help keep behavior of Python API consistent with Bash API
and vice versa, this document tries to recommend practices and styles
"from both worlds".

### DIFF
--- a/doc/api-consistency.md
+++ b/doc/api-consistency.md
@@ -1,0 +1,158 @@
+API consistency guidelines
+==========================
+
+Preupgrade Assistant provides two separate APIs for writing modules:
+Python API and Bash API.  These languages are very different, so are
+common patterns and methods of solving programming problems.
+
+Yet, we want to keep both APIs retain similar "look & feel" so that
+module developers and maintainers that will need to switch get as little
+surprise as possible.
+
+In order to help developers of both APIs prevent diverging the behaviors
+too much, this document attempts to set interface guidelines that make
+maximum use of both language's common features.
+
+
+Functions
+---------
+
+ *  Python: all functions live under `preupg` module:
+
+        preupgm.myfun()
+
+ *  Bash: all functions are prefixed by `preupgm_`:
+
+        preupgm_myfun
+
+
+Variables
+---------
+
+ *  Python: all global variables are under `preupg` module:
+
+        preupgm.MYVAR
+
+ *  Bash: all global variables are prefixed by `PREUPGM_`:
+
+        PREUPGM_MYVAR
+
+
+Arguments
+---------
+
+In general, positional parameters are preferred over key/value pairs as
+this is easier to do in Bash.
+
+In Python, only allowed types are strings, integers or lists of strings
+where a joint string would be equivalent.  It's recommended to avoid
+integers in Python since they don't have Bash equivalent (everything is
+string in Bash).
+
+
+### Scalars ###
+
+ *  Python:
+
+        preupgm.myfun('8', 'foo', 's/bar/baz/')
+
+ *  Bash: Bash only knows strings so this is obvious:
+
+        preupgm_myfun 8 foo s/bar/baz/
+
+
+### Pairs ###
+
+ *  Python:
+
+        preupgm.myfun('foo', bar=2)
+
+ *  Bash:
+
+        preupgm_myfun --bar 2 foo
+
+
+### Lists ###
+
+A long list may be passed as list of strings (or file-like objects,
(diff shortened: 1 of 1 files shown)


AloisMahdal:
TODO of this PR
=========

 *  Tell me where to put the file.  I could not find similar documents within the repo so
    I just dropped it in the root, knowing this is not the right place ;)

 *  Decide what to do with commands like `log_*`, `exit_*`, `solfmt_*`. Whether to prefrer
     single entry point with argument, or many functions with suffixes:

        log_risk_high
        log_risk_low
        ...
        solfmt_p
        solfmt_ol
        ....

    or

        risk high|low|...
        solfmt p|ol|...

    In most cases, I would prefer the latter (single point), but we sort of do the former with
    `log_*` and doing the latter has the disadvantage that it's not so pretty in Python.

 *  Decide whether `preupg_` prefix is really necessary.

 *  Of course, review, comment, discuss, schedule meetings...

api-consistency.md
```diff
@@ -0,0 +1,169 @@
...
+too much, this document attempts to set interface guidelines that make
+maximum use of both language's common features.
+
+
+Functions
+---------
+
+ *  Python: all functions live under `preupg` module:
+
+        preupg.myfun()
```
bocekm: I don't know if I understand your intentions here. We have a python package named _'preupg'_ now. You would create new module _'preupg'_ (i.e. filename _preupg.py_) within this package and use it as a python API for Preupgrade Assistant modules? That doesn't look that clean or intuitive to me.

Well, currently the python module, serving as an API for Preupgrade Assistant modules, is named _'script_api'_ and it isn't the best name either. We can change the name, that's not a problem, however we need to have it thought through. I vote for something more selfexplanatory like 'modules_api.py'. Having the name _common.sh_ for shell API is something to look at too. It can be renamed to _modules_api.sh_ for example, but that's too much off topic.

If we agree on the _preupg_ prefix, we can have the same function name in both python and shell API, i.e. _preupg_myfun_, instead of just _myfun_ as it is in the APIs now. The same with the variables. It can have name _PREUPG_MYVAR_ in both APIs.

However I don't think that the prefix is necessary. It  would compromise clarity IMHO. In python it would be like:
`from preupg import modules_api`
`modules_api.preupg_log_risk(PREUPG_RISK_LOW)`
I would prefer the nonprefix version:
`from preupg import modules_api`
`modules_api.log_risk(RISK_LOW)`

api-consistency.md
```diff
@@ -0,0 +1,169 @@
...
+            return ['baz', 'qux', 'quux']
+
+ *  Bash:
+
+        def list_foo(bar):
+            echo baz
+            echo qux
+            echo quux
+
+        is_ok() {
```
bocekm: This function shouldn't be probably here. It seems like a leftover from the **Output - boolean functions** section above.

api-consistency.md
AloisMahdal: Sorry for confusion.  Intention of this document is not to suggest rename of any modules but only to address the question of consistency between the two implementations.  I'm using the "preupg" `preupg`/`preupg_` name here mainly for historic reasons.  I did not realize it will be confusing now that the `preup` has been renamed to `preupg`.

However, I believe we both agree that at least one of them (*common.sh*, *script_api.py*) will need to be renamed anyway.  For this reason, let's keep it as t is, I'll change that as soon as the name is settled.

Regarding the prefix, that only applies to Bash API, to make up for the fact that Bash completely lacks any mechanism to control global namespace.  The idea is:

    # foo.py
    BAR = 'hey'
    def baz:
        pass

can be ported to Bash as:

     # foo.sh
     FOO_BAR = 'hey'
     foo_baz() {
         true
     }

so that following calls are equivalent:

     #!/bin/python
     import foo
     print foo.BAR
     foo.baz()

and

     #!/bin/bash
     .  foo.sh
     echo "$FOO_BAR"
     foo_baz

In our case, the "foo" is <strike>"preupg"</strike> "<to_be_decided>".  What's preserved is that:

 *  (global) constants are upper case,
 *  Bash part has virtually no chance of conflict,
 *  the name mapping is as direct as possible,
 *  all is as brief as possible,

Only thing I'm seeing as "con" for the prefixing is that at this moment things like `get_dist_native_list` would get even longer. However, I believe that should (and can, with proper API design) be addressed on the *right* part of the name, ie. by replacing eg. `s/log_error/error/`, `s/get_dist_native_list/ls_native/`...

Also note that in future users (module developers) should not be exposed to (environment) variables as most, if not all should be replaced by functions -- see my proposal #103 for details.

PR #194, opened 2016-12-15
Add specfile

bocekm:
Working with Fedora/RHEL 6. Support for RHEL-5 is to be added later.


bocekm:
Guys, please take a look at the updated spec file. It supports RHEL 5 now so it's universal (Fedora, RHEL 5/6). I added also sources, including south, pykickstart and Django, which makes 8 MB in total. Let me know if it makes sense to have them in upstream.

PR #195, opened 2016-12-15
remove component name from log message

bocekm:
- related [rhbz#1372871](https://bugzilla.redhat.com/show_bug.cgi?id=1372871)

### DIFF
--- a/common.sh
+++ b/common.sh
@@ -224,8 +224,6 @@ _log() {
 
 log_debug() {
     #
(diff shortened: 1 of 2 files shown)


bocekm:
@pirat89 Please check whether any module calls log* with component argument. If so, such module would need to be updated to remove the component argument.

bocekm:
@AloisMahdal, it is not. I'll add it.

Issue #193, opened 2016-12-14
Catch global tracebacks in preupg binary


phracek:
We have modified manual page by Pull Request #192 .
Unfortunately exit 1 is also return values caused by tracebacks.
We should detect tracebacks caused by exceptions like IndexError, ValueError and stuff like that.

bocekm:
Fixed by https://github.com/upgrades-migrations/preupgrade-assistant/commit/376a4de2e9c2ac81ac2934d3d6c6870d0289a5ab. All internal errors are caught in preupg binary and new return code is returned (30 .. settings.ReturnValues.INTERNAL_EXCEPTION).

Issue #184, opened 2016-11-30
bad error formatting when upgrade paths are missing


AloisMahdal:
When modules are missing, preupg shows message about it.  The message has broken formatting, though:

    [0 root@qeos-244 preupgrade-assistant]# preupg
    No handlers could be found for logger "preupgrade-assistant-debug"
    There were no modules found in the /usr/share/preupgrade directory.         If you would like to use this tool, you have to install some.
    [0 root@qeos-244 preupgrade-assistant]# 

Notice that the "There were no modules..." message is badly formatted.

This is from current master (65f3ac7).

bocekm:
Fixed in https://github.com/upgrades-migrations/preupgrade-assistant/commit/628f2c108fab26a8b4bbbecf073b91dd8ca90f8a and released in Preupgrade Assistant v2.3.1

Issue #165, opened 2016-11-03
Add license info to the project files


bocekm:
The license header should be present in every source file.
Also in the setup.py setup() function call - license parameter.

Issue #146, opened 2016-09-21
Script for testing BASH and python modules


phracek:
In order to support better module testing,
preupgrade-assistant should deliver a script like `preupg-test-check-script` which will check, whether
behaviour is fine or not.

phracek:
This will be handled by different way.
We should use a general wrapper around all our BASH and Python script, we are able to run them without preupg binary.

bocekm:
@AloisMahdal idea: the wrapper would also add module set API to the path. Module set API = utilities created (for example in the root of a module set) to be shared among modules to avoid duplicity.

PR #196, opened 2016-12-19
remove 'component' from generated modules and tests

bocekm:
And also from UI and man page.

This PR continues upon the already merged PR https://github.com/upgrades-migrations/preupgrade-assistant/pull/195.
Resolves: [rhbz#1372871](https://bugzilla.redhat.com/show_bug.cgi?id=1372871)

### DIFF
--- a/man/preupgrade-assistant-api.1
+++ b/man/preupgrade-assistant-api.1
@@ -91,7 +91,7 @@ Your script should also tell the administrator how risky it is to upgrade your c
 
 \fBneeds_action\fP - The test failed with the exit_fail return code but the module developer added the log_high_risk before exit_fail.
 
-\fBfail\fP - The test failed. The in-place upgrade is not recommended and marked as an EXTERME risk. 
+\fBfail\fP - The test failed. The in-place upgrade is not recommended and marked as an EXTERME risk.
 
 .SH RISK ASSESSMENT LEVELS
 The available risk assessment levels are:
@@ -110,11 +110,11 @@ The levels None, Slight and Medium change return code exit_fail to exit_needs_in
 
 There are several functions which do logging:
 
-\fBlog_{debug,info,warning,error} <component> <message>\fP
+\fBlog_{debug,info,warning,error} <message>\fP
 
 The function creates logs in the format:
 
-<SEVERITIES> <component> <TIMESTAMP> <MESSAGE>
+<SEVERITIES> <TIMESTAMP> <MESSAGE>
 
 .SH INI FILE EXAMPLE
 
--- a/preupg/ui/report/models.py
+++ b/preupg/ui/report/models.py
@@ -608,7 +608,7 @@ class TestLogMixin(object):
         create hostruns for list of hosts specified as Query in variable hosts
         """
         testlog_list = []
-        keys = ['component', 'date', 'level', 'message']
+        keys = ['date', 'level', 'message']
 
         for testlog in testlogs:
             testlog_dict = dict((key, testlog[key]) for key in keys if key in testlog)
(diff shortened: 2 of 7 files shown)


bocekm:
@AloisMahdal Another issue emerged regarding this 'easy to fix' RFE. The UI is supposed to show a history of generated assessment reports. As @pirat89 found out, it seems that with this change the UI-related processing will not be able to parse the old reports correctly (old in terms of having 'component').

If you're not able to test the UI to work correctly with both old and new reports this week, I'll drop this PR and also revert the previously merged PR #195. Then it can wait in non-master branch to be resurrected once we have more time to play with RFEs.

AloisMahdal:
Nice catch.  I'll be happy to test UI as well.

phracek:
By the way, What is the reason to remove 'component' from modules. Till now, each message is identified by component name and if it is wrong, like 'unknown' then module has to be fixed. The message should be somehow identified and we have to find out a way, how to identify it.

I would prefer to focus to important things like #193, #184, #165, #131, #146.

pirat89:
@phracek  Just my mini input, because I don't remember it completely.
I guess it is really mainly because of unkown component. It's wrong identificator because many modules can't be assigned to any specific component. If there has to be identificator, it should be module ID itself or just print info message that a module is now processed before run of the module, so it will be clear which module prints these logs.

bocekm:
@phracek This was supposed to be a quick and straightforward change, having little or no effect on the existing functionality. I agree we should focus on more important issues. The best would be to revert the previous pull request PR #195.
If we are about to keep some message identification in the log, it should be the module name, as @pirat89 suggests.

bocekm:
FYI, this affects only INFO|ERROR|WARNING messages, not RISK messages. When I look into */var/log/preupgrade/preupg.log*, the INFO|ERROR|WARNING messages are not there anyway. They are just in the result.xml/html.

phracek:
@pirat89 If there are some unknown component names, then modules should be update properly. This is only information you do not specify 'COMPONENT' in your module.

AloisMahdal:
@phracek  The reasoning behind this is in rhbz#1372871.  In short, the component part is not necessary (it's implied by module and/or message text), makes log parsing harder and is never used by any module anyway.

Also, while functional impact is not so significant, one of reasons for this is also to aid the *cleanup* goal by getting rid of unnecessary code: the idea is that by every (smaller, more focused and testable) step like this, we can shave off a little bit of complexity that makes refactoring so risky and painful.

Regarding processing old vs. new reports in UI: We could support both formats.  However I highly doubt that we have been keeping compatibility between reports, nobody ever tested this AFAIK.  Also, given that the rule is to always use latest, freshest possible assessments (with official upgrade path this is even a hard requirement), the scenario where user needs to view an old report next to new one in UI does not seem very likely to me.

man/preupgrade-assistant-api.1
```diff
@@ -110,11 +110,11 @@ The levels None, Slight and Medium change return code exit_fail to exit_needs_in
 
 There are several functions which do logging:
 
-\fBlog_{debug,info,warning,error} <component> <message>\fP
+\fBlog_{debug,info,warning,error} <message>\fP
 
 The function creates logs in the format:
 
-<SEVERITIES> <component> <TIMESTAMP> <MESSAGE>
+<SEVERITIES> <TIMESTAMP> <MESSAGE>
```
bocekm: This is actually not true, there's no timestamp anymore. The result is that the log messages are not shown in UI in the current state and will not be after this update.

pirat89:
@phracek Which component should I set for module that serves 1000 rpms? Or set of all libraries? Or when module solves troubles for just two component... etc. There is not any specific component for such modules. And we have many modules like that. Should I sest "distribution"? In that case the component doesn't have any usable information and "distribution" has same benefit as "unkown".

Agree with Lojza in this.

bocekm:
We tested this change with @AloisMahdal and it has no impact on UI.

PR #197, opened 2017-01-10
Bash API: Support multiple params for print/log funcs (#1411891)

pirat89:
Functions like log_info, log_risk_*, solution_file should process
input split into multiple parameters, because of usual constructions
in bash for printing of longer texts like this:

```bash
  $ log_info "long text" \
             "next text."
```

Previously was processed only "long text". With this fix will be
processed whole message and source code can be better formated.

Note: API for Python is kept as it is, because Python supports
      different constructions. So this difference between API
      in Bash and Python seems acceptable.


AloisMahdal:
Well... I'm not convinced that the argument behind this request is strong enough to justify change in API, let alone when it's effectively introducing an inconsistency between Bash and Python versions.

AFAIK, the use case is that a module wants to print very long log message(*). Emiting long log messages is in most cases not a good practice per se, and even when you want to do it, Bash does give you tools to do so nicely.  For example:

    msg="Strange words found:"
    msg+=" lorem"
    msg+=" ipsum"
    ...
    log_info "$msg"

or

    log_info "Strange words found: $(lsstrange)"

(*) Another use case could be printing messages from deeply nested code ... which is even worse.

So... sorry, but -1 from me.

common.sh
```diff
@@ -464,7 +465,7 @@ solution_file() {
     # solution file will be created in module directory
     # :param message: Message - string of list of strings
     #
-    echo "$1" >> "$SOLUTION_FILE"
+    echo "$@" >> "$SOLUTION_FILE"
```
phracek: This works. And what about the other functions, like exit_failed, ...? I guess, it should implement the same functionality.

phracek:
Can you describe a bit more use case? Why to not use either really long message? And from my mind. Who will read long messages?  I would prefer to keep it shorter.

common.sh
bocekm: The exit_failed accepts no argument, as I'm looking in the code. And the other functions that accept a message seem to be fixed this way too.

BTW, is there a any reason **why there's *exit_failed* in python API and *exit_fail~~ed~~* in bash API**?

bocekm:
@AloisMahdal By inconsistency you mean the number of arguments? You're right that the first _log_info_ call of the two below receives three arguments while the second _log_info_ call receives just one argument.
```
log_info "Long text" \
         "next"\
         "text."

msg="Long text"
msg+=" next"
msg+=" text"
log_info "$msg"

```
Python API expects also just one argument, even thought the style how it's written is similar to the first bash example above (with three arguments).
```
log_info("Long text"
         " next"
         " text.")
```

pirat89:
Ok, it make sense. But in that case the functions should be fixed and anywhere should appear "$@". Because now some functions accept multiple arguments and another function ignore them on the background, which make this confusing.

bocekm:
Agree. The API functions should be consistent in this matter. For example _log_info_ should call
`_log "INFO" "$1"`
and not 
`_log "INFO" "$@"`

PR #198, opened 2017-01-12
Revert "remove component name from log message"

bocekm:
Reverts upgrades-migrations/preupgrade-assistant#195
Pull request upgrades-migrations/preupgrade-assistant#196 raised questions about this change. Reverting until the concerns are resolved.


bocekm:
Both https://github.com/upgrades-migrations/preupgrade-assistant/pull/196 and https://github.com/upgrades-migrations/preupgrade-assistant/pull/195:
 - have no impact on correct functionality of the UI - reports generated before and after this change are displayed the same,
- do not impact content of preupgrade-assistant logs,
- impact only the report.xml and report.html, where it removes the _[component]_ from stdout of each module.

PR #200, opened 2017-01-13
Don't use -E sed option (not working on RHEL 5)

bocekm:
Replace with -r which has the same meaning (extended regexp)

This option actually breaks execution of every module on RHEL 5.


jmazanek:
Yes, I agree with the replacement. Apparently sed's "-E" option is undocumented and it is used for compatibility with BSD sed only.  I haven't found it anywhere in our project except the aforementioned check.sh script.

Issue #187, opened 2016-12-06
rename API modules -- finally


AloisMahdal:
Currently, we have two ports of upgrade module API:

 *  *preupg.script_api* - a Python port,

 *  *common.sh* - a Bash port.

Neither of names used here is good.

Whatever context these names might be encountered in (read: existing modules, man pages, our docs, our trackers), terms "API", "script" and "common" are all already implied. In fact, they are implied by almost *any* (Python) programming context: any Python or Bash programs can be called "scripts", Python module or a sourced Bash file already *is* an API and any API is "common" by definition.

We need to come up with better, usable and final name.

Following is what should be considered for the debate:

 *  In future, we might want to expose *outer* preupgrade-assistant API (eg. to allow people write Ansible modules).  That is, we might want to reserve `preupg` for that.

 *  Name should be brief but specific.

 *  Name should be informative, i.e. not include obvious words as mentioned above.

AloisMahdal:
One name I was thinking about is `upm`. This is based on the idea that **if** what preupgrade-assistant employs here is called "upgrade path", then the element forming that could be called "upgrade path module", ie. UPM.

That would mean:

    #!/usr/bin/python
    import preupg.upm
    preupg.upm.debug("hello")
    preupg.upm.exit('informational')

and

    #!/bin/bash
    . /usr/share/preupg/preupg_upm.sh
    preupg_upm_debug hello
    preupg_upm_exit informational

or (depending on output of discussions in #186):

    #!/bin/bash
    . /usr/share/preupg/preupg_upm.sh
    upm_debug hello
    upm_exit informational

AloisMahdal:
7am brainstorming: `upathm`, `preupm`, `preupgm`, `preupgmod`...

AloisMahdal:
We had a meetng with @bocekm and @jmazanek today.

Result: `preupgm` it is!

PR #208, opened 2017-01-23
Check for required data for crossarch migration

bocekm:
- check whether there are data in preupgrade-asssistant-*-data pkg
  that are required for a migration path from one architecture to
  another (i.e. when --dst-arch option is used)
  Related: rhbz#1414388

### DIFF
--- a/packaging/preupgrade-assistant.spec
+++ b/packaging/preupgrade-assistant.spec
@@ -42,6 +42,7 @@ Source50:       ConfigParser.py
 
 %if 0%{?rhel}
 Patch0:         preupgrade-assistant-scripts.patch
+Patch1:         crossarch_migration_support_check.patch
 %endif # RHEL
 %if 0%{?rhel} && 0%{?rhel} <= 5
 Patch100:       preupgrade-assistant-six.patch
@@ -132,6 +133,7 @@ OpenSCAP is generated automatically.
 
 %if 0%{?rhel}
 %patch0 -p1
+%patch1 -p1
 %endif # RHEL
 %if 0%{?rhel} && 0%{?rhel} <= 5
 %setup -q -n %{name}-%{version} -D -T -a 10
--- a/packaging/sources/crossarch_migration_support_check.patch
+++ b/packaging/sources/crossarch_migration_support_check.patch
@@ -1,19 +1,38 @@
 diff --git a/preupg/common.py b/preupg/common.py
-index 60286ae..55358fd 100644
+index 60286ae..7b485b3 100644
 --- a/preupg/common.py
 +++ b/preupg/common.py
-@@ -9,8 +9,10 @@ import os
+@@ -9,9 +9,11 @@ import os
  import platform
  import datetime
  import shutil
 +import sys
  from distutils import dir_util
  from preupg.utils import FileHelper, DirHelper, ProcessHelper, SystemIdentification
+-from preupg.logger import log_message
 +from preupg.exception import MissingFileInContentError
- from preupg.logger import log_message
++from preupg.logger import log_message, logger_debug, LoggerHelper
  from preupg import settings
  
-@@ -184,9 +186,23 @@ class Common(object):
+ 
+@@ -159,6 +161,16 @@ class Common(object):
+             if not os.path.exists(target_file) and os.path.exists(source_name):
+                 shutil.copyfile(source_name, target_file)
+ 
++    @staticmethod
++    def get_system_versions(modules_common_dir):
++        """Returns versions of system to be upgraded - source and destination
++        version. E.g. ["6.9", "7.3"]
++        """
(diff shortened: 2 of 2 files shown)


AloisMahdal:
My feedback based on @pirat89's output:

    [root@rhel5-32 latest]# preupgs --dst-arch model
    There are no data required by certain RHEL modules. They are expected in directory: /root/preupgrade/RHEL5_7/common/i686-model.
    The migration is most problably not supported from i686 to model architecture.
    Please see https://access.redhat.com/solutions/799813 for description of the supported migration or upgrade scenarios.

The first two lines ("There are no data required..." and  "The migration...") seem confusing. I think it would be simpler and more useful to just get to the point:

     This upgrade path does not support migration from architecture foo to bar.

Maybe mention which upgrade path we are talking about, but don't bother user by internal details like concrete paths or "certain modules".

pirat89:
As I compare both texts, the one that Lojza proposed seems better. It is short, clear and another info (like, the directory doesn't exist) can be provided by raised exception (just for example, if it is needed)

bocekm:
I think the Lojza's solution is incomplete:
1. I think using _**upgrade** path_ may confuse users even more. Maybe it's because I'm no fan of upgrade path compound. In https://github.com/upgrades-migrations/preupgrade-assistant/issues/187 we agreed on the preupgm denoting Praupgrade Assistant modules instead of Upgrade path modules.
2. What if /root/preupgrade/RHEL5_7/common/foo directory is missing (i.e. user didn't use --dst_arch option)? If I stem from the Lojza's message, then I would use "Preupgrade Assistant modules does not support migration of foo architecture.
3. The directory looked for may be missing from various reasons. That's why I chose wording "The migration is most problably not supported.." But I guess we can be like politicians, i.e. to be confident about things we are not sure about. So I guess this point may be ignored.

bocekm:
I would leave the first sentence at least as an error message in preupgrade-assistant.log, i.e.
`logger.error("There are missing data required by certain RHEL modules. They are expected in directory: %s.\n" % dir_name)`

pirat89:
> What if /root/preupgrade/RHEL5_7/common/foo directory is missing (i.e. user didn't use --dst_arch option)? If I stem from the Lojza's message, then I would use "Preupgrade Assistant modules does not support migration of foo architecture.

In that case it would be _"Preupgrade Assistant modules does not support migration from architecture foo to foo."_. This is case of migration from RHEL-5 i386, which cannot be migrated to the same architecture.

What about
```text
The migration to the new system on the foo architecture is most probably not supported
by given Preupgrade Assistant modules.
```

> I would leave the first sentence at least as an error message in preupgrade-assistant.log, i.e.
> logger.error("There are missing data required by certain RHEL modules. They are expected in directory: %s.\n" % dir_name)

What about this:
```python
logger.error("There are no available data for the  foo architecture."
                    " The '%s' directory doesn't exist.\n" % dir_name)
```

AloisMahdal:
Sorry, @bocekm. I somehow managed to miss your reply.  Here's my 3 cents:

 1. I think we should definitely agree on a term and use it
    consistently.  I admit that *upgrade path* does not come from such
    agreement; I have kind of been using it in downstream materials.
    For now, I'll use X.

    Regarding the modules, the "m" in *preupgm* correctly represents that
    the API is for **m**odules, the whole X is more than set of modules:
    it's set of modules, but also some distro data plus meta-data. At
    any rate, we don't want user to think of the X as set of objects but
    one object, no matter the internal architecture, which we might even
    change in future.

 2. I'm not sure I understand, but I imagine that at some point, preupg
    knows both architectures (whether it's provided on command line
    or some default) and can see: "does this X have the data to migrate
    from 'foo' to 'bar'? no? so then this X does not support migration
    from 'foo' to 'bar'".  I guess that's the same no matter what exactly
    is missing.

 3. I don't see how user could benefit from seeing the path. I agree
    with @pirat89 that the inclusion of directory seems like we're saying
    "you should put something in here" or "Red Hat folks should have
    put something in there" and obviously failed.

    Last, I don't think software in general should (or normally does) use
    "maybe". It seems like a weasel talk, and it's not necessary because
    the result is certain: If code finds an exception and decides to stop,
    it *will stop* and saying that a certain combination of variables is
    not supported is just stating the fact (which should be done in most
    high-level language possible). No matter if it *should* be supported
    according to some other source, like "common sense" or KB article.

bocekm:
This is now the ouput:

```
[root@localhost ~]# preupg --dst-arch ppc64 --skip-common --force
The migration from Red Hat Enterprise Linux (RHEL) 5.11 x86_64 to RHEL 7.3 ppc64 is not supported.
usage: preupg [options]

preupg: error: 
```

And to the _preupg.log_ the following is logged:
```
2017-01-30 20:13:32,835 ERROR	logger.py:47 There are no data available for the migration. The '/root/preupgrade/RHEL5_7/common/x86_64-ppc64' directory doesn't exist.
```

PR #210, opened 2017-01-31
Fix tempfile handling

AloisMahdal:
The old version created temp file (used as cache between oscap and our
log file) file "manually" as random single-letter named file under /tmp
(explicitly).  Furthermore, these files (2 of them are created with each
preupg run) are not removed afterwards.

This is unsafe and messy.  Unsafe, because adversary could easily guess
temp file name and write to it between oscap ends and file is re-read
again.  Messy because we're leaving "anonynous" random files in temp.

The fix makes use of standard Python function `tempfile.mktemp()` to
handle the tempfile creation.  While the fiename is less predictable,
this is still not completely safe: `mktemp()` still suffers from the
race-condition risk.  mktemps() from the same module does solve it but
that's not usable with ProcessHelper/FileHelper, which force us to use
file paths instead of file descriptors.

Also the fix removes the temp file after run.

I have tested this on RHEL-5 and RHEL-6; just a single quick run with
--select-rule.

### DIFF
--- a/preupg/utils.py
+++ b/preupg/utils.py
@@ -8,6 +8,7 @@ import fnmatch
 import os
 import sys
 import shutil
+import tempfile
 import mimetypes
 import platform
 import random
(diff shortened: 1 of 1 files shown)


bocekm:
I really don't like that the only way _ProcessHelper.run_subprocess_ can give you output of the command is through file. I'll focus on changing that soon - the _run_subprocess_ it should be to return the output in a variable, without the need for any temp file.
In the meantime, I agree with this fix. Thanks.

bocekm:
Created an issue https://github.com/upgrades-migrations/preupgrade-assistant/issues/211 that would get rid of the need for a temp file.

AloisMahdal:
Thanks.  Regarding the tempfile, I definitely agree. There's plenty of places in the code where simple things such as reading/writing a file or pipe are done in complex way with no advantage (or even disadvantages, as I mentioned in third paragraph of my commit message).

PR #216, opened 2017-02-03
Add new preupg-diff tool

bocekm:
preupg-diff tool compares multiple Preupgrade Assistant XML reports from which one is new with unknown issues and the others have been already analyzed.

The tool helps to find new issues that emerged in the new report by filtering out those results that are the same in the new report and at least in one of the analyzed XMLs. The trimmed report is output in XML and HTML format and named _result_diff.xml/html_.

There's also a simple test that runs the tool and checks the outcome.

Related BZ: https://bugzilla.redhat.com/show_bug.cgi?id=1418697


bocekm:
[test]

AloisMahdal:
Few first ideas:

 *  Woudn't it be better to work with tarballs rather than XMLs?  We don't have clear specification what constitutes the report: for UI it's the tarball, for r-u-t (and partly p-a) it's the /root/preupgrade directory.  Whatever it is though, the report should be always treated as inseparable unit.  Tarball fulfills this best.  Advising user to take out just the XML goes against this idea.

 *  The option to support directory instead of file path is IMHO a bit superfluous.  The feature does not add a lot: you can already use asterisk.  Dropping it would IMHO make the usage message easier.

AloisMahdal:
So I tried one scenario and it did not work: I enabled mod_perl in httpd and this issue did not appear in the diff result.

pirat89:
@bocekm see PR #246  which resolves troubles with comparing of whole solution texts. But it doesn't resolve generating of HTML files, this should be resolved too before merge.

bocekm:
To the https://github.com/upgrades-migrations/preupgrade-assistant/pull/216#issuecomment-282368093: 
- ad tarball instead of XML: it's simply too late for such a change. Sorry. The tool has gone through two rounds of beta testing by the customer who requested the tool. If I'm to change the fundametals - what's the input and possibly also the output, I would go through another round of emails whether the customer is ok with the change = another let's say month to have it here as PR again. Anyway, as you write, we don't have any specification, what exactly should be the output of Preupgrade Assistant. This tool takes one of the many generated files and I don't see any issue with that.
- ad directory option = this is the outcome of my discussion with the customer. They requested one huge master XML (merger of multiple report XMLs = not possible due to rule ID collisions). That's why I have the dir option instead. Anyway, I didn't know the asterisk works. It's a feature of shell - it expands the wildcard before the input parameters get to the tool. I agree, that the asterisk would be sufficient, but I don't think I should rely on features of user's shell.

AloisMahdal:
Regarding tarball: The issue is that if we don't have the specification what is the report, we have to somehow rely on the status quo, ie. r-u-t consuming /r/p, and UI consuming the tarballs. Introducing yet third variant seems as the opposite thing than we should do.

Regarding directory: I do not see danger in relying theon the asterisk; it's been in POSIX and I guess it's been there for ages :) Also it's currently unclear whether the directory structure should be traversed--so if you insist on the directory, this needs to be specified in help texts as well.

bocekm:
TODO: fix HTML output (html tags in solution text are dropped completely)

bocekm:
@AloisMahdal I'll drop the directory support and leave the user to use wildcards for shell or other means to provide a list of XMLs on input.

AloisMahdal:
Another issue: given two XMLs, each has only one module result: "notapplicable", then "informational" accompanied by medium risk. preupg-diff will see them as same.

pirat89:
There are still issues about missing BR tags for simple html report, but it will be fixed later, so it is not blocking for this merge.

PR #217, opened 2017-02-05
fix(2) patch for data/preassessment/scripts.txt

pirat89:
In the patched data/preassessment/scripts.txt file on RHEL system
the "df" utility is used without "-P" option for POSIX format, that
guarantees each item of the output will be printed on the one line
only without wrapping. Otherwise the output can be wrapped
and generated data will be broken or incomplete in better case. This
commit fixes use of "df".

In addition, the "find" utility of version 4.2.27 (or lesser), does
not contain option "-executable", originally used in the scripts.txt
file. This is problematic on RHEL 5 system (or just systems which
uses too old find option). Instead of this, should be used:
   "-perm /111"
which is compatible with new version of the utility too.

** this should be finally the final patch of the "scripts.txt" file **


pirat89:
Just note, that */var/cache/preupgrade/common/executable.log* **includes directories**. So I have to ask myself whether all directories would't be rather filtered out in this case.
The modules probably filter out all files that are not just of type *file* already so I prefer keep it as it is
in PA. The change can be done by modules itself separately after the functionality will be split.

PR #219, opened 2017-02-07
Update help/man page of Preupgrade Assistant

bocekm:
- add .h2m file for generating the man page from help
  message of Preupgrade Assistant

Related BZs:
- https://bugzilla.redhat.com/show_bug.cgi?id=1414359
- https://bugzilla.redhat.com/show_bug.cgi?id=1414339
- https://bugzilla.redhat.com/show_bug.cgi?id=1414392


preupg/cli.py
```diff
@@ -54,126 +54,166 @@ def __init__(self, args=None):
...
+        )
+        self.parser.add_option(
+            "-S", "--skip-common",
             action="store_true",
-            help="Skip generation of common files"
+            help="Skip generating files containing information about the"
+                 " system that are used by modules. For assessing the system"
+                 " these files are needed but they can be reused from the"
+                 " previous runs of Preupgrade Assistant."
         )
```
pirat89: The text could point out that the files are safely reusable only in case that system is in "same" state since the files were generated. That means: are installed same packages, previously unchanged configuration files are still unchanged, ... Otherwise it is recommended to create common files each run.

Note:
Probably this option would be deprecated in future, as we split logic to modules itself. Option is usable mainly for developers of the modules and preupgrade-assistant, so I think it would be mabye replaced by new directive in config file and available only under devel-mode section.

preupg/cli.py
```diff
@@ -54,126 +54,166 @@ def __init__(self, args=None):
...
+        self.parser.add_option(
+            "-f", "--force",
             action="store_true",
             default=False,
-            help="Suppress user interaction"
+            help="Suppress user interaction."
         )
         self.parser.add_option(
-            "--text",
+            "-t", "--text",
```
pirat89: Do we really want to add alternative short option, now? As the changes in options is expected "soon", I would prefer skip this, in case that will be found better use of "-t" for something else. But it is not blocking thing.

preupg/cli.py
```diff
@@ -54,126 +54,166 @@ def __init__(self, args=None):
...
-            "--verbose",
+            "-v", "--verbose",
             action="count",
             default=0,
-            help="Verbose mode"
+            help="Show more information during the assessment."
         )
         self.parser.add_option(
-            "--cleanup",
+            "-C", "--cleanup",
```
pirat89: We have already "-c" option. It could be easily overlooked that this is "-C" (just case-sensitive change). I think that it should be skipped. It is not rule that each option has to have long and short form. In addition, maybe the "--contents" option will be removed in future, so "-c" will be freed later.

preupg/cli.py
```diff
@@ -54,126 +54,166 @@ def __init__(self, args=None):
...
             action="store_true",
             default=False,
-            help="Clean data created by preupgrade-assistant"
+            help="Remove all the files created by previous runs of Preupgrade"
+                 " Assistant."
         )
         self.parser.add_option(
-            "-m",
-            "--mode",
+            "-m", "--mode",
```
pirat89: Are you sure that "-m" will not be better to use as short of "--modules" in future if we create this?

preupg/cli.py
```diff
@@ -54,126 +54,166 @@ def __init__(self, args=None):
...
+                 " its assessment by Preupgrate Assistant - migration or"
+                 " upgrade. Both modes are selected by default. This option"
+                 " may only affect behaviour of the modules - they can provide"
+                 " different results when only one mode is selected. Use one"
+                 " of these values: migrate, upgrade. It may be that modules"
+                 " behave the same no matter what mode is selected."
         )
         self.parser.add_option(
-            "--kickstart",
+            "-k", "--kickstart",
```
pirat89: The option will be probably removed in future. It should be already deprecated - I am not sure why the message is not there now. As the creation of a kickstart file is provided by the separate tool. Rather do not use "-k" here.

preupg/cli.py
```diff
@@ -54,126 +54,166 @@ def __init__(self, args=None):
...
-            help="Run a set of contents give as a comma-separated list."
+            help="Execute just a subset of modules out of a module set."
+                 " Multiple modules are to be separated by a comma." +
+                 " " * 5 +
+                 "Example:  --select-rules xccdf_preupg_rule_networking_vsftpd"
+                 "_check,xccdf_preupg_rule_networking_bind_configuration_check"
         )
         self.parser.add_option(
-            "--list-rules",
+            "-L", "--list-rules",
```
pirat89: OK, with this short option I agree :-)

preupg/cli.py
```diff
@@ -54,126 +54,166 @@ def __init__(self, args=None):
...
+            help="Provide path to all-xccdf.xml of the set of modules which is"
+                 " to be used for assesing the system. By default, if there is"
+                 " just one set in %s, Preupgrade Assistant uses that one."
+                 " Option --scan works similarly." % settings.source_dir +
+                 " " * 40 +
+                 "Example: -c /usr/share/preupgrade/RHEL5_7/all-xccdf.xml"
         )
         self.parser.add_option(
-            "--riskcheck",
+            "-R", "--riskcheck",
```
pirat89: Short option (see another comments)

preupg/cli.py
```diff
@@ -54,126 +54,166 @@ def __init__(self, args=None):
...
+                 "Example:  --select-rules xccdf_preupg_rule_networking_vsftpd"
+                 "_check,xccdf_preupg_rule_networking_bind_configuration_check"
         )
         self.parser.add_option(
-            "--list-rules",
+            "-L", "--list-rules",
             action="store_true",
             default=False,
-            help="List available contents"
+            help="List all the modules available within a module set."
```
pirat89: Really it will be in the one module set or for all module sets? Should be then specified/mentioned use of --scan or -c to choose correct module set if there exists more modules? (in case it will print all dinf rules of all modules sets, it will not be probably required)

preupg/cli.py
```diff
@@ -54,126 +54,166 @@ def __init__(self, args=None):
...
         )
         self.parser.add_option(
-            "--version",
-            action="store_true",
-            default=False,
-            help="Show Preupgrade Assistant version"
-        )
-        self.parser.add_option(
-            "--dst-arch",
+            "-D", "--dst-arch",
```
pirat89: Short option.

preupg/cli.py
```diff
@@ -54,126 +54,166 @@ def __init__(self, args=None):
...
+                "Specify an architecture of the system to be migrate to."
+                " Available option are: %s. Use of the option is expected on"
+                " 32-bit systems as by the release of RHEL 7, 32-bit hardware"
+                " support has been dropped."
+                % ", ".join(settings.migration_options)
+            )
         )
         self.parser.add_option(
-            "--old-report-style",
+            "-o", "--old-report-style",
```
pirat89: -o is usually used for output file. As somewhere was requested fueature to specify where should be report file - or kickstart? - stored, it will not be intuitive. I prefer keep here only long form of the option.

preupg/cli.py
bocekm: Agree. I added some "warning".

preupg/cli.py
bocekm: Ok, I'll remove the shorter versions. They can be introduced later during the CLI overhaul.

preupg/cli.py
bocekm: I'd like to have the shorter versions for most of the options and Capital short options are quite common. Anyway let's remove it for now as the CLI will be changed in the near future.

preupg/cli.py
bocekm: It is supposed to list modules of just a single module set (_"a module set"_ means just a single one).
However when there are multiple module sets, -s option is to be used. However it doesn't work, I'll create an issue.

preupg/cli.py
bocekm: This short option has been there before this pull request.

bocekm:
@pirat89 the requested changes have been done.

preupg/cli.py
```diff
@@ -1,11 +1,11 @@
 # -*- coding: utf-8 -*-
 
 from __future__ import print_function, unicode_literals
-#import argparse
 import optparse
```
pirat89: In future, we would use argparse if it is available and fallback to optparse otherwise. I created an issue #223 for such change.

preupg/cli.py
pirat89: Ah, that's true. My mistake.

preupg/cli.py
```diff
@@ -54,126 +55,162 @@ def __init__(self, args=None):
 
     def add_args(self):
         self.parser.add_option(
-            "--skip-common",
+            "-h", "--help",
+            action="help",
+            help="Show help message and exit."
+        )
+        self.parser.add_option(
+            "-S", "--skip-common",
```
pirat89: I missed that short option probably yet.

bocekm: I thought I would leave this one because it's used quite often.

preupg/cli.py
bocekm: I'm not sure if the argparse is backwards compatible. If yes, then why not. We may try it one time. However currently, neither RHEL 5 nor RHEL 6 have argparse available so this RFE is for quite a distant future.

preupg/cli.py
bocekm: Issue https://github.com/upgrades-migrations/preupgrade-assistant/issues/225 created.

preupg/cli.py
pirat89: Ok, so this one we will keep.

bocekm:
[test]

PR #220, opened 2017-02-08
RHEL5: fix generating of report file

pirat89:
Previously there were use of python3 "print" function rewrite to use of python2 print function. But the rewrite changed the output, as there was used stdout instead of stderr sometimes; sometimes was produced string with two newlines ending instead of one newline character. This is fixed now.
 
The old openscap doesn't provide separately stderr and stdout output of each module and both merges into the stdout. Stderr is usually empty (or more precisely, contains whitespaces only).

To be more compatible with possible tests written on RHEL 6 system, merge stdout and stderr outputs in the results.xml file together as stderr and stdout remove.

I am not sure about possible outputs of OpenSCAP for all situations, so solution seems more ugly than it has to be.

### DIFF
--- a/packaging/sources/preupgrade-assistant-xml-logs.patch
+++ b/packaging/sources/preupgrade-assistant-xml-logs.patch
@@ -32,7 +32,7 @@ index 1a9047c..7223bf4 100644
          """
          #Filter all rule-result in TestResult
          changed_fields = []
-+        self.fix_stderr(self) # it is ugly but command below too
++        self.fix_stderr() # it is ugly but command below too
          self.remove_empty_check_import()
          inplace_dict = {
              0: ReportHelper.upd_inspection,
--- a/packaging/sources/preupgrade-assistant-xml-logs.patch
+++ b/packaging/sources/preupgrade-assistant-xml-logs.patch
@@ -1,0 +1,38 @@
+diff --git a/preupg/report_parser.py b/preupg/report_parser.py
+index 1a9047c..7223bf4 100644
+--- a/preupg/report_parser.py
++++ b/preupg/report_parser.py
+@@ -198,6 +198,25 @@ class ReportParser(object):
+ 
+         self.write_xml()
+ 
++    def fix_stderr(self):
++        """The function transform stdout to stderr."""
++        for rule in self.get_all_result_rules():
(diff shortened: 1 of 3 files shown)


bocekm:
[test]

packaging/sources/preupgrade-assistant-xml-logs.patch
```diff
@@ -0,0 +1,38 @@
...
+ 
+         self.write_xml()
+ 
++    def fix_stderr(self):
++        """The function transform stdout to stderr."""
++        for rule in self.get_all_result_rules():
++        # Filter all check-import=stdout which are empty and remove them
++            for check in self.get_nodes(rule, "check"):
++                cim_stdout = [x for x in self.get_nodes(check, "check-import")
++                                if x.get('import-name') == "stdout"]
```
bocekm: It makes no sense to get a list of all the check-import=stdout, when later on you work just with the first element. However keep it as is as this is low priority and temporary patch which will be removed soon.

packaging/sources/preupgrade-assistant-xml-logs.patch
```diff
@@ -0,0 +1,38 @@
...
++                cim_stderr = [x for x in self.get_nodes(check, "check-import")
++                                if x.get('import-name') == "stderr"]
++                if not cim_stdout or cim_stdout[0].text is None or cim_stdout[0].text == "":
++                    continue
++
++                # unexpected, but just to be sure..
++                if cim_stderr:
++                    if cim_stderr[0].text is not None and cim_stderr[0].text != "":
++                        cim_stdout[0].text +="\n"+cim_stderr[0].text
++                    remove_node(check, cim_stderr[0])
```
bocekm: Removing the stderr node here assumes there's just one stderr node - which will probably be always the case. However the condition is "_if cim_stderr:_" and _cim_stderr_ is a list, so the logic is strange - if there's multiple stderr nodes, just the first one would be removed.
Also I don't like much removing empty nodes. There may be utilities which expect the nodes to be there (for example preupg-dif :). I know, it'd be a fault of the utility to not be robust enough, if it collapses when the expected node's not there, but anyway.
However, keep the code as is due to what I wrote in the comment above.

PR #222, opened 2017-02-08
Minor cleanup

bocekm:
Exit instead of continue if XML template not found.
Make exceptions more verbose.
Another minor cleanup:
  - code wrap on 80 chars
  - classmethod to staticmethod, where classmethod not needed
  - fixed typos
  - removed unused imports or variables

### DIFF
--- a/preupg/xmlgen/compose.py
+++ b/preupg/xmlgen/compose.py
@@ -212,9 +212,9 @@ class ComposeXML(object):
 
             return prefix, tree_key
 
-        for f in sorted(six.iterkeys(group_tree),
-                        key=lambda tree_key:
-                        get_sorting_key_for_tree(group_tree, tree_key)):
+        def sort_key(t_key):
+            return get_sorting_key_for_tree(group_tree, t_key)
+        for f in sorted(six.iterkeys(group_tree), key=sort_key):
             t = group_tree[f]
             tree, subgroups = t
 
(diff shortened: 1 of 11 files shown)


preupg/xmlgen/compose.py
```diff
@@ -144,61 +157,71 @@ def perform_autoqa(cls, path_prefix, group_tree):
...
+                             [-1].text)
             except:
                 pass
 
             return prefix, tree_key
 
-        for f in sorted(six.iterkeys(group_tree), key=lambda tree_key: get_sorting_key_for_tree(group_tree, tree_key)):
+        for f in sorted(six.iterkeys(group_tree),
+                        key=lambda tree_key:
+                        get_sorting_key_for_tree(group_tree, tree_key)):
```
pirat89: plase define the lambda function above instead of this two-lined split. Like:
```python
sort_key=lambda t_key: get_sorting_key_for_tree(group_tree, t_key)
for f in sorted(six.iterkeys(group_tree), key=sort_key)):
```

bocekm: Ok, I'll change it but I'm getting PEP-8 E731 warning to not to assign lambda expression to a variable. And I kind of agree that there's a better/more comprehensive way to write the code - using def for defining function. All right, it's two lines instead of one, but who cares if it makes the code easier to comprehend.

bocekm:
@pirat89 see my reply for your comment. Anyway I changed the code you wanted to change :)
Also I fixed formatting of a message of one of the exceptions.

PR #228, opened 2017-02-09
Improve handling of exceptions

bocekm:
- fixed internal Exceptions (super() does not work with py2.4)
- when there's an Exception or where previosly there was just sys.exit(), Preupgrade Assistant now returns a new code 29
- by default, tracebacks are not shown, however by using --debug option, the traceback is shown
- return code 22 changed from RISK_CLEANUP_KICKSTART, which made little sense (riskcheck + cleanup + kickstart options can't be used at the same time), to INVALID_CLI_OPTION
- add new return code 30 for USER_ABORT - situations where user decides to abort the assessment (by Ctrl+C or answering No on a question)
- when printing an error message, replace regular print() with sys.stderr.write()


bocekm:
[test]

bin/preupg
```diff
@@ -14,19 +15,18 @@ def main():
...
-        except Exception as ex:
-            cli.parser.error(str(ex))
-    return ret
+    try:
+        return_code = app.run()
+        logger_debug.debug("preupg returned '%s'" % return_code)
+        return return_code
+    except KeyboardInterrupt:
+        return settings.ReturnValues.USER_ABORT
+    except Exception as ex:
```
pirat89: this line is not covered by patch for python-2.4. It seems for now that it is the only one problem, just according to usual run of "preupg". I will look at it closer yet today.

bocekm: thanks. fixed.

bocekm:
[test]

preupg/xmlgen/compose.py
```diff
@@ -339,8 +335,8 @@ def get_xml_tree():
         try:
             target_tree = ElementTree.parse(template_file).getroot()
         except IOError:
-            sys.exit('Problem with reading %s file'
-                     % settings.xccdf_template)
+            raise IOError('Error: Problem with reading %s file'
+                          % settings.xccdf_template)
```
pirat89: This is OK.

preupg/xmlgen/xml_utils.py
```diff
@@ -309,11 +306,11 @@ def create_xml_from_ini(self, main):
         }
         for key in self.keys:
             if 'check_script' not in key:
-                raise MissingTagsIniFileError
+                raise MissingTagsIniFileError(tags="check_script", ini_file=main)
```
pirat89: Note: Here is the place, where missing 'check_script' and 'solution' tags are detected. (my comment about  above content_title)?

bin/preupg
bocekm: Good point. I'll add printing a newline.

preupg/application.py
```diff
@@ -489,13 +493,10 @@ def scan_system(self):
...
-                lines = [i.rstrip() for i in FileHelper.get_file_content(os.path.join(self.assessment_dir,
-                                                                                      self.conf.mode),
-                                                                         'rb',
-                                                                         method=True)]
-            except IOError:
-                return
+            lines = [i.rstrip() for i in FileHelper.get_file_content(os.path.join(self.assessment_dir,
+                                                                                  self.conf.mode),
+                                                                     'rb',
+                                                                     method=True)]
```
bocekm: TL;DR: Basically, this was just a poor error handling.

If user uses the --mode option, only the appropriate modules are supposed to be run (either for upgrade or migration). And for that two files exist - _/root/preupgrade/RHEL5_7/{migrate,upgrate}_ - holding a list of modules to run. However if there is any error with reading this file, the scan_system function just returns None and no further assessment happens. User (and then also developer who would troubleshoot this) is left with no information whatsoever what went wrong.

preupg/xmlgen/compose.py
bocekm: IOError does not append \n to the end of the message. The final \n is added in /bin/preupg when any Exception is caught.

preupg/xmlgen/xml_utils.py
bocekm: Yes, here and also in the _check_recommended_fields_ function (called three lines below). It's quite a hack (or let's say quick patching of whatever problem that you can see across the whole code of Preupgrade Assistant).
I fixed it in PR 230: https://github.com/upgrades-migrations/preupgrade-assistant/pull/230/files#diff-ae96a7b8baaa7ea98cc9b74552c59221L311

preupg/xmlgen/oscap_group_xml.py
```diff
@@ -55,33 +55,28 @@ def find_all_ini(self):
...
+            filehander = codecs.open(file_name, 'r', encoding=settings.defenc)
+            config.readfp(filehander)
+            fields = {}
+            if config.has_section('premigrate'):
+                section = 'premigrate'
+            else:
+                section = 'preupgrade'
+            for option in config.options(section):
+                fields[option] = config.get(section, option)
+            self.loaded[file_name] = [fields]
```
bocekm: The message you printed is related to _MissingTagsIniFileError_, which was not caught here. It must be caused by a different change. I'll try to find it out.

bocekm: I've tried to replicate your ouput and I'm not successful:
```
$ tools/preupg-xccdf-compose /<path>/RHEL6_7
Error: Tag(s) content_title missing in the /<path>/RHEL6_7-results/backup/bacula/module.ini file
```
Please make sure you are running the correct code from this branch.

bocekm: There is a code in _check_recommended_fields_ that makes no sense, that causes this issue. It does not raise the correct exception when some of the required fields ['content_title', 'check_script', 'solution'] is missing AND recommended field 'applies_to' is missing in INI.
I've fixed it in https://github.com/upgrades-migrations/preupgrade-assistant/pull/230/files#diff-9eae4399e4b933991c4b81856022426dL169

bocekm:
Issues raised by @pirat89 fixed.

AloisMahdal:
Yay! `\o/`

PR #230, opened 2017-02-10
xmlgen refactoring

bocekm:
- simplify loading and parsing INI files
- rename strangely named variables
- fix checking whether required fields are in INI
- update unit tests and patch accordingly

### DIFF
--- a/preupg/xmlgen/xml_utils.py
+++ b/preupg/xmlgen/xml_utils.py
@@ -10,7 +10,7 @@ from preupg.utils import SystemIdentification, FileHelper
 from preupg import settings
 from preupg.xmlgen import xml_tags
 from preupg.xmlgen.script_utils import ModuleHelper
-from preupg.exception import MissingTagsIniFileError, EmptyTagGroupXMLError
+from preupg.exception import EmptyTagGroupXMLError
 
 
 def get_full_xml_tag(dirname):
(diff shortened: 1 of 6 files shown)


bocekm:
Rebased with master.

preupg/xmlgen/xml_utils.py
```diff
@@ -209,16 +207,17 @@ def check_script_modification(self, key, k):
         self.update_values_list(self.rule, "{" + k + "}", key[k])
 
     def prepare_sections(self):
-        """The function prepares all tags needed for generation group.xml file."""
-        for main, self.keys in self.ini_files.items():
-            if main.endswith("group.ini"):
+        """The function prepares all tags needed for generation group.xml file.
+        """
+        for ini_filepath, ini_file_content in iter(self.ini_files.items()):
```
AloisMahdal: Isn't the `iter()` call redundant?  (Just checked with python 2.7 and it seems to behave the same)

pirat89: You are right, that behaviour inside *for* loop is same and probably redundant in most cases, as the `self.ini_files.items()` returns list of items, which is naturaly iterated by *for* loop.

bocekm: Agree, I used http://legacy.python.org/dev/peps/pep-0469/ as a base for this change, but in case of _for_ cycle, the _iter_ is redundant. As @pirat89 mentioned, I used this style on many places I guess it would be easier to make a PR to deal with just usage of this redundant _iter_ :)

PR #231, opened 2017-02-10
Basic code cleanup

bocekm:
- replaced using six library by python built-in functions
  - the code without six is still python 2.x and 3.x compatible
- remove unused imports, commented out code
- wrap code to 80 chars
- functionality is kept the same

### DIFF
--- a/preupg/application.py
+++ b/preupg/application.py
@@ -544,13 +544,13 @@ class Application(object):
             path = self.openscap_helper.get_default_html_result_path()
 
         report_dict = {
-            0: settings.message.format(path),
-            1: settings.message.format(path),
-            2: 'We found some critical issues. In-place upgrade is not'
-               ' advised.\n' + "Read the file {0} for more details.".
+            0: settings.risks_found_warning.format(path),
+            1: settings.risks_found_warning.format(path),
+            2: 'We have found some critical issues. In-place upgrade or migration is not advised.\n' +
+            "Read the file {0} for more details.".
             format(path),
-            3: 'We found some error issues. In-place upgrade is not advised'
-               '.\n' + "Read the file {0} for more details.".format(path)
+            3: 'We have found some error issues. In-place upgrade or migration is not advised.\n' +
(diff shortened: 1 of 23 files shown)


pirat89:
I gues that some commits can be dropped. E.g. for python-2.4 patches.

bocekm:
Waiting for https://github.com/upgrades-migrations/preupgrade-assistant/pull/258 to be merged to master. Then I'll do a rebase and thus remove deprecated patches.

preupg/kickstart/plugins/partitioning.py
```diff
@@ -182,21 +202,29 @@ def get_volume_info(filename, first_index, second_index):
     def get_partition_layout(self, lsblk, vgs, lvdisplay):
         """
         Returns dictionary with partition and realname and size
-        :param filename:  filename with partition_layout in /root/preupgrade/kickstart directory
+        :param filename:  filename with partition_layout in
+                          /root/preupgrade/kickstart directory
```
pirat89: is it really up-to-date description? see header of func.

preupg/application.py
```diff
@@ -35,10 +34,12 @@
 
 
 def fault_repr(self):
-    """monkey patching Fault's repr method so newlines are actually interpreted"""
+    """monkey patching Fault's repr method so newlines are actually interpreted
+    """
```
bocekm: That's against documentation convention (https://www.python.org/dev/peps/pep-0257/)

bocekm: I'll change it to
```
"""Monkey patching Fault's repr method to interpret newlines."""
```

preupg/kickstart/plugins/partitioning.py
bocekm: Most probably isn't :) I changed it to have each line less then 80 chars, but you're right, I'll update the content to reflect the current params.

AloisMahdal:
Better late than never --- I looked at the diff and all seems OK.

I also ran few downstream tests with builds from your private COPR and all passed.

Issue #124, opened 2016-08-02
weird behaviour around dirname of set of modules


pirat89:
Functionality of PA is really bounded to name of directory, like RHEL6_7. For different names - format of name - in better cases print error message and ends. However, for another cases it will not print any troubles and another processing and results are really wrong.

E.g.:
- RHEL6_7-results - processing ends OK, however some replace_holder strings in results.html are not replaced (like *_SOLUTION_TEXT)
- 6_7 - contents end as not applicable but any message about name of directory, which causes this
  .... etc

At least PA have to check correctly if everything is OK and print relevant error message and end with error when there is used name of directory, which affects significantly another processing and required format of name must be documented.

However, some behaviour mentioned above shouldn't be really affected by the name of directory.

phracek:
I see the point. But module dirname should contain at least numbers.
And w/o numbers it should work too. I will incorporate the change soon.

phracek:
@AloisMahdal @pirat89 What about to have only dirname or whatever user specifies and inside dirname a file which specifies an upgrade path?
But definitely we should have specified an upgrade path.

AloisMahdal:
IMO: Anything that defines the upgrade path should be inside the folder; the folder *name* should not bear any information whatsoever.

For meta-data, a file or set of files seems an obvious solution.  The format is not really important as it's wholly managed by automated tools (we don't expect path devels/users to edit or even see the data; this also means we can always change the format later) so I'd go for something dead simple.

I hope I have answered your question.

phracek:
Yeah, I will implemented in the next release cycle and make sense to me.

phracek:
This issue will be solved by PR #148

bocekm:
The change is implemented in https://github.com/upgrades-migrations/preupgrade-assistant/pull/233.
Resolves rhbz: https://bugzilla.redhat.com/show_bug.cgi?id=1412539.

AloisMahdal:
@bocekm, a little correction: rhbz#1412539, is a "meta-bug" and suggests several changes (some even in modules). There is no direct downstream counterpart to this change; I planned to file each bug once we agree which parts we want to do.

bocekm:
Another two PRs that addresses this issue: PR https://github.com/upgrades-migrations/preupgrade-assistant/issues/280 and https://github.com/upgrades-migrations/preupgrade-assistant/issues/286

bocekm:
[...]

PR #233, opened 2017-02-14
Get rid of dirname dependency

MajoDurco:
Solved #124 issue.
**Idea:**
Module set name can be arbitrary now. Instead, each module set needs to have _properties.ini_ file in its directory root. Example of the INI content:
```
[preupgrade-assistant-modules]
srcMajorVersion=6
dstMajorVersion=7
```
This PR also simplifies structure of the testing module set (used by unit tests) and removes unused test files.


bocekm:
Next week I'll provide a little training on how to write unit tests for Preupgrade Assistant. Then you can fix the failing tests/write new ones.

AloisMahdal:
About testing: current behavior adds complexity to some widely shared parts of our downstream testing code (eg. upgrade path generator).  After this change, I will be able to significantly simplify these parts; if most of the test suite does not go bonkers, it should be OK.

pirat89:
I suggest to rebase the branch when currently aprooved PRs are merged. Then I will test functionality.

bocekm:
Rebase done. Now the unit tests need to be fixed.

preupg/settings.py
```diff
@@ -43,6 +43,15 @@
...
 log_dir = "/var/log/preupgrade"
 
+# file with meta info
+properties_ini = "properties.ini"
+# section inside ini file
+section_name = "preupgrade-assistant-modules"
+# key which specify source major OS version
+src_version_key ="srcMajorVersion"
+# key to specify destination major OS version
+dest_version_key ="destMajorVersion"
```
pirat89: Use dstMajorVersion istead of "dest". When you use already "src", "dst" is the best equivalent.

bocekm: Expected behaviour after talking to @pirat89:

- When you encounter _properties.ini_ in the place where it is expected (_/usr/share/preupgrade/<module_set_dir>/properties.ini_) and it has no "preupgrade-assistant-modules" section print out warning, something like _"Found 'properties.ini' in \<dir\>. If it's supposed to be ini file for a module set, it needs to contain 'preupgrade-assistant-modules' section."_

- When the _properties.ini_ file has _"preupgrade-assistant-modules"_  section but it has either of the two options missing (_"srcMajorVersion"_ or _"destMajorVersion"_) print out an error message and stop Preupgrade Assistant execution.

AloisMahdal: Why warnings? The first part looks like if we're trying to negotiate the format with user.

We don't (and should not) provide any flexibility here. We have trivial file with prescribed mandatory format and placement. If user makes a typo, there's literally 2-3 obvious places to look at.

IMO all this needs is:

    try:
        parse properties.ini
        read [preupgrade-assistant-modules].srcMajorVersion
        read [preupgrade-assistant-modules].dstMajorVersion
    except AnyError:
        crash with message such as "missing mandatory meta-data key: srcMajorVersion"

Ie. make it easy for user to get the gist of what's wrong with the set (meta-data) and give them something to grep (srcMajorVersion).

preupg/settings.py
```diff
@@ -43,6 +43,15 @@
...
+
 # preupg log file
 preupg_log = os.path.join(log_dir, "preupg.log")
```
AloisMahdal: One comment line per LOC is just way, way too many comments.  Please don't add comments unless you really have to; especially don't add comments that just say the same thing as the code says.

My advice would be to just remove all comments from this hunk. Or even better:

... because, wait ... why are there even "settings" for this? There's no need for this; these values are not really subject to change. IMO they should just be hard-coded into the hunk below, otherwise this just destroys readability (you have to look in *other file* just to see a "variable" that is never, ever supposed to change).

tests/generated_results/inplace_risk_test.xml
```diff
@@ -187,4 +187,4 @@
     <ns0:score maximum="100.000000" system="urn:xccdf:scoring:default">49.081196</ns0:score>
     <ns0:score maximum="70.000000" system="urn:xccdf:scoring:flat">40.000000</ns0:score>
   </ns0:TestResult>
-</ns0:Benchmark>
+</ns0:Benchmark>
```
AloisMahdal: This is a whitespace noise.

Note that the UNIX convention is that all text files should have newline at the end. Whoever committed this needs to fix their editor, otherwise the next person touching the file (with proper editor) will see exactly opposite hunk.

preupg/settings.py
bocekm: I have no problem with @AloisMahdal approach. The probability that somebody would create a directory for something else than modules in the default place for modules (/usr/share/preupg/) and put properties.ini with some arbitrary content in it is so close to zero, that we can ignore it.

preupg/utils.py
```diff
@@ -411,27 +411,33 @@ def get_convertors():
...
+        """
+        >>> get_valid_scenario('/dir1/dir2/file.xml')
+        'dir2'
+        >>> get_valid_scenario('/dir1/dir2/dir3/')
+        'dir3'
+        >>> get_valid_scenario('/dir1/dir2/dir3')
+        'dir2'
+        # doesn't check directory at the end of path, is considered as a file
+        """
+        return os.path.basename(os.path.split(scenario_path)[0])
```
bocekm: Use _os.path.dirname_ instead of _os.path.split(scenario_path)[0]_. It is equivalent and more clear.

Consider renaming the method to _get_module_set_dirname_. The current name does not hint much what it's supposed to do.

bocekm: Also, before the get_valid_scenario provided some kind of validation, that the directory is indeed a module set. The validation was through the name of the directory - it was now replaced with properties.ini.
So, there should be a function that checks whether the provided path to all-xccdf.xml points to a valid module set, i.e. that it contains properties.ini. Something like:
```
def is_module_set_valid(all_xccdf_path):
  # pseudocode below
  if os.path.dirname(all_xccdf_path) contains properties.ini:
    return True
  else:
    return False
```

preupg/settings.py
pirat89: When you say it in this way, I am OK with that too.

bocekm:
[test]

tests/test_preupg.py
```diff
@@ -273,95 +274,125 @@ def test_solution_www_bold_tag(self):
...
-    def setUp(self):
-        self.temp_dir = tempfile.mkdtemp()
-
-    def tearDown(self):
-        shutil.rmtree(self.temp_dir)
-
+    '''
+    Test get_scenario method, to get directory with modules
+    '''
     def test_correct_content_scenario(self):
```
bocekm: Rename to use the current terminology in which content=>module, scenario=module set.

tests/test_preupg.py
```diff
@@ -273,95 +274,125 @@ def test_solution_www_bold_tag(self):
...
-class TestPreupgradePrefix(base.TestCase):
-    def setUp(self):
-        settings.prefix = 'preupgrade'
-
-    def test_correct_prefix(self):
-        version = SystemIdentification.get_assessment_version('FOOBAR6_7')
+        self.assertEqual(a.get_scenario(), 'Modules')
+
+
+class TestValidScenario(base.TestCase):
```
bocekm: Rename to TestValidModuleSet

tests/test_preupg.py
```diff
@@ -273,95 +274,125 @@ def test_solution_www_bold_tag(self):
 
 
 class TestScenario(base.TestCase):
```
bocekm: Rename to TestModuleSet

preupg/utils.py
```diff
@@ -565,6 +543,14 @@ def get_preupg_config_file(full_path, key, section="preupgrade-assistant"):
             if config.has_option(section, key):
                 return config.get(section, key)
 
+    @staticmethod
+    def has_ini_file_section(ini_file_path, section):
```
MajoDurco: Method only parse INI files so what if configuration file is not INI file?

preupg/utils.py
```diff
@@ -862,3 +848,74 @@ def run_generate(self, xml_file, html_file, old_style=False):
         return ret_val
 
 
+class ModulSetUtils(object):
+
+    @staticmethod
+    def check_property(ini_file_path, key, section):
```
MajoDurco: Basically the same question as above. Maybe I can rename it as:
```def get_ini_key_value(...)```

File: preupg/utils.py (hunk 8 above)
Comment (bocekm) [inline preupg/utils.py]: I've spoken with @MajoDurco directly. The outcome is that the default/standard configuration file in python has the INI format and is parsed by ConfigParser library. When someone mentions _python_ 
and _config_, IMO most of the time they mean [this python standard format](https://docs.python.org/2/library/configparser.html).

File: preupg/application.py [hunk 10]
```diff
@@ -489,13 +517,11 @@ def scan_system(self):
...
             return ReturnValues.SCENARIO
         if not self.conf.contents:
-            version = SystemIdentification.get_assessment_version(
-                self.conf.scan)
-            if version is None:
-                log_message("Your scan is in a wrong format %s." % version,
-                            level=logging.ERROR)
-                log_message("It should be like 'RHEL6_7' for upgrade from"
-                            " RHEL 6->7.", level=logging.ERROR)
+            version = None
```
bocekm: Fixed in https://github.com/upgrades-migrations/preupgrade-assistant/pull/233/commits/107a550a8ec5aeb8ddcbd5e59fef0d6a99e3e0a2

preupg/settings.py
```diff
@@ -43,6 +43,9 @@
 cache_dir = "/var/cache/preupgrade"
 log_dir = "/var/log/preupgrade"
 
+# file with meta info
```
bocekm: https://github.com/upgrades-migrations/preupgrade-assistant/pull/233/commits/107a550a8ec5aeb8ddcbd5e59fef0d6a99e3e0a2

preupg/utils.py
bocekm: Fixed in https://github.com/upgrades-migrations/preupgrade-assistant/pull/233/commits/107a550a8ec5aeb8ddcbd5e59fef0d6a99e3e0a2

preupg/utils.py
```diff
@@ -862,3 +848,74 @@ def run_generate(self, xml_file, html_file, old_style=False):
         return ret_val
 
 
+class ModulSetUtils(object):
```
bocekm: Fixed in https://github.com/upgrades-migrations/preupgrade-assistant/pull/233/commits/107a550a8ec5aeb8ddcbd5e59fef0d6a99e3e0a2

preupg/utils.py
bocekm: Fixed in https://github.com/upgrades-migrations/preupgrade-assistant/pull/233/commits/107a550a8ec5aeb8ddcbd5e59fef0d6a99e3e0a2

preupg/utils.py
```diff
@@ -862,3 +848,74 @@ def run_generate(self, xml_file, html_file, old_style=False):
...
+        key_value = ConfigHelper.get_preupg_config_file(ini_file_path,
+                                                        key, section)
+        if not key_value:
+            raise EnvironmentError("Key {0} inside {1} section\n"
+                                   "is missing in {2}"
+                                   .format(key, section, ini_file_path))
+        return key_value
+
+    @staticmethod
+    def get_assessment_version(module_set_path):
```
bocekm: Fixed in https://github.com/upgrades-migrations/preupgrade-assistant/pull/233/commits/107a550a8ec5aeb8ddcbd5e59fef0d6a99e3e0a2

preupg/utils.py
```diff
@@ -862,3 +848,74 @@ def run_generate(self, xml_file, html_file, old_style=False):
...
+        section_name = "preupgrade-assistant-modules"
+        if not os.path.isdir(module_set_path):
+            module_set_path = os.path.join(settings.source_dir,
+                                           module_set_path)
+        properties_file = os.path.join(module_set_path,
+                                       settings.properties_ini)
+        if not os.path.isfile(properties_file):
+            raise EnvironmentError("{0} file not found inside {1} directory"
+                                   .format(settings.properties_ini,
+                                           os.path.dirname(properties_file)))
```
bocekm: Fixed in https://github.com/upgrades-migrations/preupgrade-assistant/pull/233/commits/107a550a8ec5aeb8ddcbd5e59fef0d6a99e3e0a2

tests/test_preupg.py
bocekm: Fixed in https://github.com/upgrades-migrations/preupgrade-assistant/pull/233/commits/107a550a8ec5aeb8ddcbd5e59fef0d6a99e3e0a2

tests/test_preupg.py
bocekm: Fixed in https://github.com/upgrades-migrations/preupgrade-assistant/pull/233/commits/107a550a8ec5aeb8ddcbd5e59fef0d6a99e3e0a2

tests/test_preupg.py
bocekm: Fixed in https://github.com/upgrades-migrations/preupgrade-assistant/pull/233/commits/107a550a8ec5aeb8ddcbd5e59fef0d6a99e3e0a2

tests/generated_results/inplace_risk_test.xml
bocekm: Fixed in https://github.com/upgrades-migrations/preupgrade-assistant/commit/194ab3ecbeb1ff37d2787648f732f830eda3b1b5

AloisMahdal:
Removing qa_ack as this was already merged.

AloisMahdal:
qa_ack: re-verify (as master) with latest version of downstream test suite (`ipu tier pa=U pa-*`)

AloisMahdal:
Re-ran test suite with `preupgrade-assistant-2.5.0-1.201710201148Z.master.el6_8`; passed.

PR #235, opened 2017-02-14
mitigate security issues in kickstart

bocekm:
- it may pose a security risk to put the root password into the
  kickstart - use a placeholder instead. User will need to fill
  in the password first before using the kickstart
- set the default hashing algorithm for user passwords to the
  strong sha-512.

### DIFF
--- a/data/templates/preupgrade.ks
+++ b/data/templates/preupgrade.ks
@@ -10,8 +10,10 @@ halt
 timezone Etc/UTC
 # System language
 lang en_US
+# Root password
+rootpw <password_placeholder>
 # System authorization information
-auth  --useshadow  --passalgo=sha512
+auth  --enableshadow --passalgo=sha512
 # enable firstboot
 firstboot --enable
 # SELinux configuration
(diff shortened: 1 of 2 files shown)


jmazanek:
From my experience, it is customary to use some kind of weak "dummy" root password in this phase of server migration, which is then replaced with the "real" one before the system is commissioned for production, so the security risk is IMO not that great in this case. System is at this point usually disconnected from the outside network and what matters is smooth automatic installation.
But it seems, that the others have already made their mind about it, so I don't object.

bocekm:
@jmazanek the security risk I originally meant is this scenario: 1. admin has a system to migrate with a strong root password which shouldn't be exposed and the system uses "only" md5 hash for passwords, 2. admin generates kickstart, that contains the root password md5 hash., 3. admin copies the kickstart to any unsecure place (e.g. files a bug or sends an email with the kickstart as an attachment) and there you have it, the password has leaked.

bocekm:
By the way, before this update the root password was taken from the original kickstart (the one used during the system installation). So if any password appears in the generated kickstart, it might be the current root password or it might not :)

PR #236, opened 2017-02-14
Autopart instead of custom partitioning in kickstart

bocekm:
- set autopart to let the new system set up partitions automatically
- comment out custom partitioning commands
Related:
- https://bugzilla.redhat.com/show_bug.cgi?id=1414408
- https://bugzilla.redhat.com/show_bug.cgi?id=1414341

### DIFF
--- a/preupg/kickstart/application.py
+++ b/preupg/kickstart/application.py
@@ -134,6 +134,13 @@ class KickstartGenerator(object):
                 return None
         return ksparser
 
+    def set_autopart(self):
+        """Letting kickstart to do the partitioning automatically is the safe
+        option (as opposed to guessing the correct partitioning parameters for
+        the new system).
+        """
+        self.ks.handler.autopart(autopart=True)
+
     def delete_obsolete_issues(self):
         """ Remove obsolete items which does not exist on RHEL-7 anymore"""
         self.ks.handler.bootloader.location = None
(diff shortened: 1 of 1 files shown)


bocekm:
What may currently get into the generated kickstart is bootloader **--driveorder**:
`bootloader --append="console=tty0 console=ttyS0,115200 rd_NO_PLYMOUTH" --driveorder="vda"`
I'm not sure whether that may be a problem if the new system does not recognize the drive identifier, in this case _vda_.
Please advise, if you have the knowledge related to this. Otherwise, I'll keep it as is.

jmazanek:
I haven't tested it myself, so I can't tell if in case of missing valid drive, there is "auto-magically" used some default, but I doubt it. Obviously the "vda" drive won't be recognised in case of migration to the physical machine. Since we can't predict what will be the type of the target system ( virtual or physical), we should ensure, that the user will review this part of kickstart and edit it as needed.

bocekm:
I've created a beaker job that should install RHEL 7 using the kickstart with _--driveorder="vda"_.
However it's been an hour and the installation hasn't started yet. We'll see tomorrow.
https://beaker.engineering.redhat.com/recipes/3520101#installation

AloisMahdal:
I encountered the `--driveorder=vda` problem during test dat dry run.  If I remember correctly, it only caused warning in anaconda log, but did not halt the automated installation.

bocekm:
I confirm that the _driveorder_ does not cause issues. The beaker installation was successful even though the only available drive was **sda**.

PR #237, opened 2017-02-16
kickstart: fix postmigrate template for kickstart

pirat89:
- unpack tarball with '--selinux' option to keep original selinux
  context
- modified and added log messages
- remove creating of ".diff" files when restore cleanconf configs
  - WHY:
    - the files would be overwritten when some configuration files
      have same names but different paths
    - the log file was messy because of trailing diff output
    - admin is able to create diff of the modified files
      very simply
- backup original files with suffix .preupg_save instead of .rpmsave
  which could be rewritten during installation/update of new
  packages

- fix paths for
- create symlinks for other important directories
  - added: noauto_postupgrade.d, postmigrate.d
    (we can discuss creation of symlinks for all 'main' directories
     in /root/preupgrade)

### DIFF
--- a/data/templates/postmigrate.sh
+++ b/data/templates/postmigrate.sh
@@ -26,7 +26,7 @@ do
     SAVE_PATH="${ABS_PATH}.preupg_save"
     if [[ -f "${ABS_PATH}" ]]
     then
-        echo "INFO: Back up new configuration file to '${SAVE_PATH}'." >> ${PREUPGRADE_LOG}
+        echo "INFO: Backing up the clean system configuration file '${ABS_PATH}' to '${SAVE_PATH}'." >> ${PREUPGRADE_LOG}
         cp -a ${ABS_PATH} ${SAVE_PATH}
     fi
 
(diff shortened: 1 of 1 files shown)


data/templates/postmigrate.sh
```diff
@@ -5,38 +5,48 @@ cd ${TEMP_DIR}
...
     ABS_PATH=${file:1}
-    SAVE_PATH="${ABS_PATH}.rpmsave"
-    DIFF_FILENAME="$(basename ${file}).diff"
-    echo "Comparing file ${file} against ${SAVE_PATH}" >> ${PREUPGRADE_LOG}
-    echo "Diff output is stored in ${TEMP_DIR}/${DIFF_FILENAME}" >> ${PREUPGRADE_LOG}
-    [[ -f ${ABS_PATH} ]] && \
+    SAVE_PATH="${ABS_PATH}.preupg_save"
+    if [[ -f "${ABS_PATH}" ]]
+    then
+        echo "INFO: Back up new configuration file to '${SAVE_PATH}'." >> ${PREUPGRADE_LOG}
```
bocekm: I wouldn't call the conf file that you are backing up as "new". When I looked at the log I was confused a bit. And present continuous tense may be more suitable too.
I suggest something like:
"INFO: Backing up the clean system configuration file '${ABS_PATH}' to '${SAVE_PATH}'."

An then for the restoration:
"INFO: Restoring configuration file of the migrated system '${ABS_PATH}'."

pirat89:
Agree, it sounds better.

PR #240, opened 2017-02-21
remove the BEL characters when reloads XML file

pirat89:
The OpenSCAP put whole stdout/stderr into the result.xml directly
so we are not able to handle these outputs. In case that the BEL
character is produced by some utility (it happened already), parsing
of XML file fails.

NOTE: Some XML libraries handle it already, but not the one, used
      by the PA, mainly on older systems

### DIFF
--- a/preupg/report_parser.py
+++ b/preupg/report_parser.py
@@ -106,7 +106,11 @@ class ReportParser(object):
         content = FileHelper.get_file_content(self.path, 'rb', False, False)
         if not content:
             return None
-        self.target_tree = ElementTree.fromstring(content)
+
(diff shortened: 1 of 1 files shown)


AloisMahdal:
Why just `BEL`? For example, XML 1.0 allows only `0x09`, `0x0a` and `0x0d`.  This code could be nicer, perhaps?

    xml10invalid = bytearray([n for n in xrange(0,31) if n not in [9,10,13]])
    self.target_tree = ElementTree.fromstring(content.translate(None, xml10invalid))

(and you could leave the comment out since it's obvious...

PR #248, opened 2017-02-28
Add possibility to change data_dir with env variable

MajoDurco:
**Problem:**
    Preupgrade Assistant tools in this moment depend on hardcoded paths
    from setting.py including data_dir, which points to
    '/usr/share/preupgrade/data'. When PA isn't installed on the system
    and developer wants to test a new code(e.g. run xccdf-compose) he
    needs to edit setting.py in order to change data_dir variable.
**Solution:**
    To be able to override this data_dir variable with environment
    variable "PREUPG_DATA_DIR" instead of changing setting.py

### DIFF
--- a/preupg/settings.py
+++ b/preupg/settings.py
@@ -36,6 +36,9 @@ share_dir = "/usr/share"
 source_dir = os.path.join(share_dir, prefix)
 
 data_dir = os.path.join(source_dir, "data")
+# set data_dir value based on env variable PREUPG_DATA_DIR
+# if variable is not set doesn't change the value
+data_dir = os.getenv('PREUPG_DATA_DIR', data_dir)
 
 # file where the lock file stored
 lock_file = "/var/run/preupgrade.pid"


AloisMahdal:
But why?

bocekm:
@AloisMahdal to be able to run the tools (like xccdf-compose, preupg-diff) without the need to install them (either by rpm or setup.py install). It's mainly a feature for developers.

AloisMahdal:
@MajoDurco Please consider writing the purpose and real effect (ie. more specific, not just about some internal variable) into the commit message.

(In case you're not familiar with git/github, you can just `git commit --amend`, edit the message and `git push --force` over the same branch; github will not be confused by this).

bocekm:
I agree with @AloisMahdal, the commit/PR description should have been more informational. @MajoDurco please update the commit message (_commit --amend_ + _push --force_).

To remedy the missing information, here's the rationale for the change:
**Current situation:**
Preupgrade Assistant and its tools depend on the paths provided by settings.py. All of them have been hardcoded, including the _data_dir_, which points to _"/usr/share/preupgrade/data"_. However, the data are not present in this directory unless the Preupgrade Assistant is installed on the system. The workaround for developers who want to test their new & fresh code is to edit settings.py.
**Solution:**
This change introduces an environment variable which, if set, overrides the hardcoded _"/usr/share/preupgrade/data"_. Developer then needs to set this environment variable (for example in ~/.bashrc) instead of changing temporarily the settings.py.

PR #253, opened 2017-03-07
Update xccdf_template.xml

bocekm:
Merged PR https://github.com/upgrades-migrations/preupgrade-assistant/pull/216 introduced an incorrect update to the xccdf_template.xml - _html_ namespace was removed from \<p\> tags. However, internally, OpenSCAP validates the _all_xccdf.xml_ generated from a module set and this validation fails. It does expect \<html:*\> tags, i.e. \<html:p\> was correct in the template.

### DIFF
--- a/data/templates/xccdf_template.xml
+++ b/data/templates/xccdf_template.xml
@@ -2,9 +2,9 @@
   <status date="${CURRENT_DATE}">draft</status>
   <title>Preupgrade Assistant analysis report</title>
   <description>
-    <p>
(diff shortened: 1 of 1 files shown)


bocekm:
[test]

AloisMahdal:
I updated the file manually on last downstream build and the new template works fine.

AloisMahdal:
Just note that this does not affect on module sets built with a version before #216 when the regression was introduced.

PR #255, opened 2017-03-09
Proof of concept: attach_file() function

AloisMahdal:
Already wrote a (partial) test downstream.

To do:

 *  Decide what to do (fail, warn, silently overwrite...) with repeated
    submissions.

 *  Afterwards, consider merging storage places with hooks in deploy_hook()
    (seems more logical than hooks/MODNAME/...)

 *  Probably document ability to use slash in NAME, meaning a subdir
    (works now).

 *  Python version.

 *  Solve linking.


pirat89:
It seems reasonable to me as PoC. Just remember, that the directory has to be part of the tarball with result. See the *preupgrade_dirs* inside *preupg/settings.py*

pirat89:
--test--

pirat89:
--test--

common.sh
```diff
@@ -695,6 +703,59 @@ add_pkg_to_kickstart() {
...
+        "")
+            full_dst="$_MODULE_STORAGE/files/${src##*/}"
+            ;;
+        *)
+            full_dst="$_MODULE_STORAGE/files/$dst"
+            ;;
+    esac
+    log_debug "attach_file(): full_dst='$full_dst'"
+    mkdir -p "${full_dst%/*}"
+    cp "$src" "$full_dst"
```
pirat89: use rather `cp -a` to preserve mode, selinux context, ...

common.sh
```diff
@@ -194,6 +194,14 @@ else
...
 fi
 
+#
+# Module-generated files storage path
+#
+# This should be used as root for things like hooks and
+# attachments.  Don't save files directly here, create
+# a subdirectory for each type of file.
+#
+_MODULE_STORAGE=$VALUE_TMP_PREUPGRADE/storage/$MODULE_PATH
```
pirat89: The directory has to be packed into the tarball. See `preupgrade_dirs` inside `setting.py`.

AloisMahdal:
Thanks to the review.

Rebased with current master and implemented (almost all) requested changes (incl. @pirat89).

> - _Afterwards, consider merging storage places with hooks in deploy_hook() (seems more logical than hooks/MODNAME/...)_
>
> Yeah, that makes sense. Something like:
>
>     $VALUE_TMP_PREUPGRADE/storage/$MODULE_PATH/{files,hooks}

This will be done in future PR, though; I'd like to keep this one purely additive.


> - _Probably document ability to use slash in NAME, meaning a subdir (works now)._
>
> Yeah, add info there about this possibility.

Added, including few examples to make it more clear.

> Consider adding some security check to not let user save the file outside the
> `$_MODULE_STORAGE/files` directory, e.g. by using "NAME=../../../../etc/passwd"

Added "jailbreak" checking function `_is_rooted_in()`, which returns true if a path (computed from DST given by user) "belongs" to another path (storage path), even after canonicalization.  This is called before touching the destination and if it fails, `log_error()` and `exit_error()` are called.

> - _Python version._
>
> Yes please, create the python version.

Definitely; I'd like to settle on implementation mechanics first.

> - _Solve linking._

The aim is to enable mentioning the file in solution text in a way that preupg will be able to convert it to a valid link.

I actually mentioned it in `attach_file()` docstring, but the way I described it there is a bit ahead of itself.  It  already assumes usage of Markdown via functions ported from [mdfmt.sh][1].  This assumption comes from discussions we had about implementing Markdown support for solution texts.  However, these discussions have not been finished, so for now I'll just skip the "linking" part and concentrate merely on the ability to attach files per se.

Hence, I'm removing that part of docstring trom next force-push.

  [1]: https://github.com/AloisMahdal/shellfu/blob/master/src/include-sh/mdfmt.sh

common.sh
AloisMahdal: Added.

common.sh
bocekm: Agree. See https://github.com/upgrades-migrations/preupgrade-assistant/pull/271 for details.

AloisMahdal: Already done in next version, though. (GitHub should now be marking this thread as outdated).

common.sh
```diff
@@ -737,5 +802,34 @@ deploy_hook() {
...
+    local path=$1       # PATH
+    local jail=$2       # JAIL
+    local cpath         # canonical PATH
+    local cjail         # canonical JAIL
+    local subpath       # sub-directory part of PATH
+    cpath=$(readlink -m "$path")
+    cjail=$(readlink -m "$jail")
+    subpath=${cpath#$cjail/}
+    test "$subpath" == "$cpath" && {
+        log_error "path points outside jail: $cpath not in $cjail"
```
bocekm: You're calling log_error twice, one right after the other, with very similar message. I would leave just one log_error, probably the one on line 739 - this way the caller is the one to decide what to log/print. Anyway, since the log_error message is shown to the customer in the report I'm not sure if they would understand the problem. I'm adding Jirka to review it.

AloisMahdal: Fair enough, I'll remove the one from `_is_rooted_in()` -- it will be a "silent" function.

AloisMahdal:
Rebased with current master and removed the double error call.

@jhornice, the call that remains is the outer one, eg.

      destination NAME points outside storage: ../../some/path

I have also realized that having `$full_dst` in the error message makes it less helpful since the path is prefixed by a lot of "junk" we are adding outside user's control.  So I changed it to `$dst` -- which is exactly what user passes and is documented as `NAME`.

Note that the context is that user has called `attach_file()` with optional NAME (they want to submit file but under a specific name).

pirat89:
--test--

PR #256, opened 2017-03-13
Remove creation of admin and user reports

MajoDurco:
result-admin.xml/html, result-user.xml/html are reduntant in PA because
the final result is the same as in the default result.xml/html

### DIFF
--- a/doc/README
+++ b/doc/README
@@ -39,14 +39,6 @@ The files are:
 result.html - a file with the final migration assessment report in a human-readable
          form (the functionality is only listed)
 result.xml - a file with the final migration assessment report in a machine-readable form
-result-admin.html - a file with the final migration assessment report in a human-readable
-         form (the functionality is only listed) for administrators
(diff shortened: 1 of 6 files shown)


bocekm:
Related BZ: https://bugzilla.redhat.com/show_bug.cgi?id=1414322

AloisMahdal:
IIRC, there was also an element in UI that mentioned these reports; I guess chage might be necessary there as well. (But I might be wrong...)

AloisMahdal:
Ack for next cycle; this will be really easy to add to in downstream suite (preupg/result-files test), plus we'll verify the UI, which should also pretty straightforward.

bocekm:
@MajoDurco, I guess the code here https://github.com/upgrades-migrations/preupgrade-assistant/blob/remove_admin_report/preupg/xmlgen/xml_utils.py#L321 can be removed too.

AloisMahdal:
`qa_verified`: Already added to downstream test suite and verified as part of previous release.

PR #258, opened 2017-04-25
Remove RHEL 5 code

bocekm:
- support for RHEL 5 has been dropped


bocekm:
a) Man page updated.
b) _preupg/ui/report/models.py_ - The mention of RHEL 5 should stay here because the UI server should IMO still support gathering reports even from remote RHEL 5 machines.
c) _preupg/ui/utils/distrib.py_ - The code in is basically dead. It's not called anywhere. RHEL 5 is mentioned there as part of a hardcoded path - some ugly temporary code. I don't think it's worth doing a cleanup of UI though.

PR #259, opened 2017-04-28
Fix generating Preupgrade Assistant modules

bocekm:
When generating module set for Preupgrade Assistant using _tools/preupg-xccdf-compose_, XML parse errors were occuring.
Error example:
```
Encountered a parse error in file /home/mbocek/.../RHEL5_7-results/services/pki/pki-ca/group.xml.
Details: mismatched tag: line 42, column 8
```
It was due to unclosed HTML tags in generated group.xmls.

### DIFF
--- a/preupg/xmlgen/xml_tags.py
+++ b/preupg/xmlgen/xml_tags.py
@@ -58,8 +58,8 @@ CONFIG_SECTION = """
             File(s) affected:
             <ul>
             {config_file}
-            <ul>
-        <p>
+            </ul>
+        </p>
 """
 RULE_SECTION_VALUE_IMPORT = """\t\t<check-import import-name="stderr"/>"""
 


AloisMahdal:
Were the warnings the only symptom or is there a behavior change?

bocekm:
@AloisMahdal, no, because this piece of HTML goes only into <description> of a rule:
```
    <Rule id="xccdf_preupg_rule_services_openssh_..." selected="true">
      <title>OpenSSH ...</title>
      <description>
        This content has a aim to convert ...
        <p>
            File(s) affected:
            <ul>
            <li>/etc/ssh/sshd.sh</li>
            <ul>
        <p>
      </description>
      <fixtext>_services_openssh_sshd_SOLUTION_MSG_TEXT</fixtext>
      <check system="http://open-scap.org/page/SCE">
        <check-import import-name="stdout" />
        ...
        <check-content-ref href="openssh-sshd.sh" />
      </check>
    </Rule>
```

bocekm:
This is actually still causing problems, when running the system assessment:
```
001/140 ...running (Configuration files to be reviewed)OpenSCAP Error: File '/root/preupgrade-assistant-contents-users-master/RHEL6_7-results/all-xccdf.xml' line 866: Element '{http://checklists.nist.gov/xccdf/1.2}p': This element is not expected. Expected is one of ( {http://www.w3.org/1999/xhtml}*, {http://checklists.nist.gov/xccdf/1.2}sub ).
```
Created PR to fix it: https://github.com/upgrades-migrations/preupgrade-assistant/pull/263
I introduced the problem here: https://github.com/upgrades-migrations/preupgrade-assistant/commit/82564095fac73bb74728d2bc54a9839e16515ef4

PR #263, opened 2017-05-12
Revert back removed xhtml namespace

bocekm:
- without it, the default ns0 is used, which is openscap namespace
  and this namespace doesn't define html tags - that leads to crash
  during the xml parsing

### DIFF
--- a/preupg/xmlgen/xml_tags.py
+++ b/preupg/xmlgen/xml_tags.py
@@ -53,13 +53,14 @@ FIX = """<fix """+PLATFORM+""" system="urn:xccdf:fix:script:{script_type}">
 {solution}
          </fix>"""
 FIX_TEXT = """<fixtext>{solution_text}</fixtext>"""
+
 CONFIG_SECTION = """
-        <p>
+        <xhtml:p>
             File(s) affected:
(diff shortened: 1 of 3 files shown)


bocekm:
This is actually still causing problems, when running the system assessment:
```
001/140 ...running (Configuration files to be reviewed)OpenSCAP Error: File '/root/preupgrade-assistant-contents-users-master/RHEL6_7-results/all-xccdf.xml' line 866: Element '{http://checklists.nist.gov/xccdf/1.2}p': This element is not expected. Expected is one of ( {http://www.w3.org/1999/xhtml}*, {http://checklists.nist.gov/xccdf/1.2}sub ).
```
I introduced the problem here: upgrades-migrations/preupgrade-assistant@8256409

bocekm:
It's caused by this change: https://github.com/upgrades-migrations/preupgrade-assistant/commit/817fe90d50f022db36db5a38b96faef11c2c23ad#diff-cd214281061a49f12a09f465b45bc87bR395
The line should be:
```
for third_party, content in list_contents(dir_name).items():
```
I'll create new PR for that, it's not related to this one.

Issue #225, opened 2017-02-08
--list-rules does not work with -s option


bocekm:
There comes a problem when there are more module sets than one (say RHEL5_7 and my_custom5_7) in /usr/share/preupgrade. The using the --list-rules, Preupgrade Assistant suggests to use -s option to choose which module set we'd like to use. However if -s is used, nothing else happens. Still the same suggestion to use -s option, even though I've used it.

bocekm:
I can see @AloisMahdal has already reported this in Bugzilla: https://bugzilla.redhat.com/show_bug.cgi?id=1418509

bocekm:
Fixed by PR https://github.com/upgrades-migrations/preupgrade-assistant/pull/269

PR #269, opened 2017-05-19
Fix listing rules in case of multiple modules sets

bocekm:
- When more than one module set was available, --list-rules option was not working
- Related:
  - https://bugzilla.redhat.com/show_bug.cgi?id=1418509
  - Issue https://github.com/upgrades-migrations/preupgrade-assistant/issues/225

### DIFF
--- a/preupg/application.py
+++ b/preupg/application.py
@@ -103,7 +103,6 @@ class Application(object):
         self.tar_ball_name = None
         self.third_party = ""
         self.assessment_dir = None
-        self.list_scans = []
 
     def _add_report_log_file(self):
         """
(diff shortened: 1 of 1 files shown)


preupg/application.py
```diff
@@ -581,31 +580,23 @@ def _set_devel_mode(self):
...
             self._devel_mode = 0
 
-    def _check_available_contents(self):
-        cnt = 0
-        is_dir = lambda x: os.path.isdir(os.path.join(self.conf.source_dir, x))
-        dirs = os.listdir(self.conf.source_dir)
-        self.list_scans = []
-        for dir_name in filter(is_dir, dirs):
-            if SystemIdentification.get_assessment_version(dir_name):
-                self.conf.scan = dir_name
```
bocekm: Here, the function was changing the self.conf.scan, which is basically available from anywhere of the code Preupgrade Assistant. I don't think function with _check_ in its name should be doing such a nastiness. So instead of changing it, I'm returning it.

Also this for loop does basically the same as the already existing _list_contents()_ function. So I used it.

preupg/application.py
```diff
@@ -623,14 +614,15 @@ def run(self):
...
-                return ret_val
+            self.conf.scan = self._get_default_module_set()
+            if not self.conf.scan:
+                return ReturnValues.SCENARIO
 
         if self.conf.list_rules:
-            ret_val = self._check_available_contents()
-            if int(ret_val) != 0:
-                return ret_val
+            if not self.conf.scan:
```
bocekm: The line 622 is basically the only line needed for the fix. The other changes were done to make the code a bit more sane.

AloisMahdal:
Checked with downstream (copr) build.  If multiple sets are installed:

    preupg --list-rules

requires us to select one using `-s`.  If that is added, list of rules is produced correctly.

bocekm:
@AloisMahdal I believe that is the expected behaviour.

PR #273, opened 2017-07-12
Allow using -u option without -r

bocekm:
Fixes running Preupgrade Assistant when running it with just a _-u_ option:
`preupg -u http://hostname:8099/submit`
This is a basic command that used to work and is documented on many places. So this is a high priority one.

Resolves: [rhbz#1391968](https://bugzilla.redhat.com/show_bug.cgi?id=1391968)

### DIFF
--- a/preupg/application.py
+++ b/preupg/application.py
@@ -629,7 +629,7 @@ class Application(object):
             log_message('\n'.join(rules))
             return 0
 
-        if self.conf.upload:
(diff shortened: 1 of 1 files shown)


AloisMahdal:
Tested with devel build: works fine. If I add `-u URL`, preupg runs and result is uploaded afterwards. If I add `-u URL -r PATH`, specified result tarball is uploaded immediately.

PR #282, opened 2017-07-31
Dir name dependent tags in generated xml files with xccdf-compose

MajoDurco:
When run:
```
preupg-xccdf-compose RHEL6_7
```
tag generated in group.xml looks like:
```xml
<value>SCENARIO/backup/bacula</value>  
```
but when run:
```
preupg-xccdf-compose moduleDir
```
tag looks like:
```xml
<value>SCENARIO/root/moduleDir-results/backup/bacula</value>
```
This concerns multiple tags through xml files not only one showed above.

### DIFF
--- a/preupg/xmlgen/compose.py
+++ b/preupg/xmlgen/compose.py
@@ -77,8 +77,45 @@ class XCCDFCompose(object):
 class ComposeXML(object):
 
     @staticmethod
-    def collect_group_xmls(source_dir, content=None, level=0,
+    def collect_group_xmls(root_module_dir, source_dir, content=None, level=0,
                            generate_from_ini=True):
+        """
+        Find group.xml file recursively through all module directories
+        and modules. Collect data from each of them into dictionary.
+
+        @parma {str} root_module_dir - directory where all modules are stored
+        @param {str} source_dir - directory path for processing
+        @param {str} content - module result e.g. failed,needs_action,pass,...
+        @param {int} level - indicate the depth of recursive call in directory
+        @param {bool} generate_from_ini - True if xccdf-compose tool is used
+
+        @return {dict} - structure is file based, keys are top level module
+        directories, values are tuples which consist of 2 elements:
+            [0] - XML data for top level dir itself
+            [1] - dict of all modules inside [0] directory
+        @example
+        from file structure:
+            - RHEL6_7
+                - services (modules directory)
+                    - group.xml
+                    - group.ini
+                    - httpd (module)
+                        - group.xml
+                        - ...
+                    - tomcat (module)
+                        - group.xml
+                        - ...
+                - drivers (modules directory)
+                - ...
+        result is:
+            {services: tuple(<Element {services}>, {
+                httpd: tuple(<Element {httpd}>, {}),
+                tomcat: tuple(<Element {tomcat}>, {}),
+                ...
+            )},
+            drivers: tuple(<Element {drivers}>, {...}),
+            ...}
+        """
         ret = {}
 
         for dirname in os.listdir(source_dir):
@@ -100,7 +137,7 @@ class ComposeXML(object):
                            " in %s. Please check if it is really missing."
                            % new_dir)
             if ini_files and generate_from_ini:
-                oscap_group = OscapGroupXml(new_dir)
+                oscap_group = OscapGroupXml(root_module_dir, new_dir)
                 oscap_group.write_xml()
                 return_list = oscap_group.collect_group_xmls()
                 ComposeXML.perform_autoqa(new_dir, return_list)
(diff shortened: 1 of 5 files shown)


preupg/xmlgen/compose.py
```diff
@@ -77,8 +77,44 @@ def get_compose_dir_name(self):
...
+                        - group.xml
+                        - ...
+                - drivers (modules directory)
+                - ...
+        result is:
+            {services: tuple(<Element {services}>, {
+                httpd: tuple(<Element {httpd}>, {}),
+                tomcat: tuple(<Element {tomcat}>, {}),
+                ...
+            )},
```
bocekm: The indentation makes the result confusing.
Also this one line 114 has should be "}),".
I would make the dictionary more clearly indented:
```
            {
                services: tuple(
                    <Element {services}>,
                    {
                        httpd: tuple(<Element {httpd}>, {}),
                        tomcat: tuple(<Element {tomcat}>, {}),
                        ...
                    }
                ),
                drivers: tuple(
                    <Element {drivers}>,
                    {...}
                ),
                ...
            }
```

preupg/xmlgen/compose.py
```diff
@@ -96,11 +132,11 @@ def collect_group_xmls(source_dir, content=None, level=0,
                 directories = [x for x in os.listdir(new_dir)
                                if not os.path.isdir(os.path.join(new_dir, x))]
                 if not directories and 'postupgrade.d' not in dirname:
-                    print ("WARNING: It seems that group.ini file is missing"
-                           " in %s. Please check if it is really missing."
-                           % new_dir)
+                    print("WARNING: It seems that group.ini file is missing"
+                          " in %s. Please check if it is really missing."
+                          % new_dir)
```
bocekm: Simplify the warning to: 
`"group.ini file is missing in %s."`
and use _log_message_ function for it, with _level=logging.WARNING_.

preupg/xmlgen/compose.py
```diff
@@ -110,12 +146,12 @@ def collect_group_xmls(source_dir, content=None, level=0,
...
-                                level=level + 1,
-                                generate_from_ini=generate_from_ini))
+                                ComposeXML.collect_group_xmls(
+                                    module_set_dir, new_dir,
+                                    generate_from_ini=generate_from_ini))
             except ParseError as e:
-                print ("Encountered a parse error in file ", group_file_path,
-                       " details: ", e)
+                print("Encountered a parse error in file ", group_file_path,
+                      " details: ", e)
```
bocekm: Use log_message function for this error message, with level=logging.ERROR. Then it should stop execution. I think sys.exit(1) can be sufficient.

preupg/xmlgen/xml_utils.py
```diff
@@ -13,24 +12,43 @@
...
-    for index, dir_name in enumerate(dirname.split(os.path.sep)):
-        if re.match(r'\D+(\d)_(\D*)(\d)-results', dir_name, re.I):
-            found = index
-    main_dir = dirname.split(os.path.sep)[found + 1:]
-    return main_dir
+def module_path_from_root_dir(dirname, module_set_dir):
+    """
+    Remove module_set_dir from dirname path
+
+    @param {str} dir_name - directory of specific method or method directory
```
bocekm: method -> module

preupg/xmlgen/xml_utils.py
```diff
@@ -13,24 +12,43 @@
...
 
 
 class XmlUtils(object):
     """Class generate a XML from xml_tags and loaded INI file"""
-    def __init__(self, dir_name, ini_files):
+
+    def __init__(self, module_set_dir, module_dir, ini_files):
+        """
+        @param {str} module_set_dir - directory where all modules are stored
+        @param {str} module_dir - directory of specific method or method
```
bocekm: method -> module

preupg/xmlgen/compose.py
```diff
@@ -77,8 +77,44 @@ def get_compose_dir_name(self):
 class ComposeXML(object):
 
     @staticmethod
```
bocekm: Make this method a non-static so you don't need to pass the _module_set_dir_ and _generate_from_ini_ to the method.

AloisMahdal:
qa_ack: I guess manual verification and test suite re-run should suffice

AloisMahdal:
> `qa_ack`: I guess manual verification and test suite re-run should suffice.

Manual verification is not possible with normally accessible builds, since shipped version does not allow the use case from OP.  We could hunt for the build OP used but it's not worth; it's enough to do test suite re-run.

Which I just did with downstream-packaged 2.5.0 and it passed.

PR #283, opened 2017-07-31
Check for properties.ini when xccdf-compose is used

MajoDurco:


MajoDurco:
Same change made in #287. Closing duplicate.

Issue #286, opened 2017-08-02
preupgrade-content-creator is broken


AloisMahdal:
As of current master (using internal build `2.4.4-1.201707271409Z.el6_8`), preupg-content-creator is completely broken; it will die at first question, no matter what I type in (even if it's exactly the same string as the error message suggests---a hack that does work with released version).

    # preupg-content-creator
    Specify the upgrade path (like RHEL6_7) where the module will be stored: FOO
    The scenario 'FOO' is not valid.
    It has to be like RHEL6_7 or CentOS7_RHEL7.
    # preupg-content-creator
    Specify the upgrade path (like RHEL6_7) where the module will be stored: RHEL6_7
    The scenario 'RHEL6_7' is not valid.
    It has to be like RHEL6_7 or CentOS7_RHEL7.
    # preupg-content-creator
    Specify the upgrade path (like RHEL6_7) where the module will be stored: CentOS7_RHEL7 
    The scenario 'CentOS7_RHEL7' is not valid.
    It has to be like RHEL6_7 or CentOS7_RHEL7.
    # 

(I suspect PR #233.)

AloisMahdal:
Note that this blocks downstream testing of other PRs that are based on current master; I have no way to create mock modules now.

MajoDurco:
Yes, you are right. It's caused with #233. I'll try to fix it ASAP.

MajoDurco:
#287

PR #287, opened 2017-08-03
Fix getting right module set directory

MajoDurco:
Based on #286. 

Preupgrade-content-creator was looking for a specific upgrade path,
replace this behavior with no dependency on directory name and also add
properties.ini into new created module set directory.

In compose.py check for existing properties.ini file in exitsting
module set directory instead of in directory with '-results' postfix.

### DIFF
--- a/preupg/creator/ui_helper.py
+++ b/preupg/creator/ui_helper.py
@@ -12,11 +12,9 @@ import sys
 from distutils.util import strtobool
 from preupg.utils import FileHelper, ModuleSetUtils
 from preupg.creator import settings
+from preupg import settings as preupgSettings
 
 from preupg.settings import content_file as ALL_XCCDF_XML
-
-section = 'preupgrade'
-PROPERTIES_FILE_NAME = 'properties.ini'
 
 
 def get_user_input(message, default_yes=True, any_input=False):
@@ -129,13 +127,15 @@ class UIHelper(object):
         self.content_path = os.path.join(self.get_group_name(),
                                          self.get_content_name())
 
-    def properties_ini(self):
-        """ If properties.ini file doesnt exist ask user for OS versions """
+    def ask_about_properties_ini(self):
+        """ If properties.ini file doesnt exist ask user for OS versions,
+        otherwise don't ask anything
+        """
         self.properties_ini_path = os.path.join(
             self.get_upgrade_path(),
-            PROPERTIES_FILE_NAME)
+            preupgSettings.properties_ini)
         if not self.properties_ini_exists():
-            self.get_properties_ini_versions()  # ask user for versions
+            self.get_properties_ini_versions()  # ask user forn  versions
 
     def properties_ini_exists(self):
         if self.properties_ini_path:
@@ -229,10 +229,10 @@ class UIHelper(object):
         :return:
         """
         config = ConfigParser.RawConfigParser()
-        config.add_section(section)
+        config.add_section(preupgSettings.prefix)
         for key, val in iter(self.content_dict.items()):
             if val is not None:
-                config.set(section, key, val)
+                config.set(preupgSettings.prefix, key, val)
 
         self.content_ini = self.get_content_name() + '.ini'
         ini_path = os.path.join(self.get_content_path(), self.get_content_ini_file())
@@ -264,8 +264,8 @@ class UIHelper(object):
         :return:
         """
         config = ConfigParser.RawConfigParser()
-        config.add_section(section)
-        config.set(section, 'group_title', 'Title for %s ' % self.get_group_name())
+        config.add_section(preupgSettings.prefix)
+        config.set(preupgSettings.prefix, 'group_title', 'Title for %s ' % self.get_group_name())
 
         file_name = 'group.ini'
         group_ini = os.path.join(self.get_upgrade_path(),
@@ -328,7 +328,7 @@ class UIHelper(object):
             if self.specify_upgrade_path() is None:
                 return 1
 
-            self.properties_ini()
+            self.ask_about_properties_ini()
 
             self.get_content_info()
             if self.refresh_content:
--- a/preupg/creator/ui_helper.py
+++ b/preupg/creator/ui_helper.py
@@ -127,7 +127,7 @@ class UIHelper(object):
         self.content_path = os.path.join(self.get_group_name(),
                                          self.get_content_name())
 
-    def ask_about_properties_ini(self):
+    def get_properties_ini_info(self):
         """ If properties.ini file doesnt exist ask user for OS versions,
         otherwise don't ask anything
         """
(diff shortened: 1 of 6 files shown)


preupg/creator/ui_helper.py
```diff
@@ -195,6 +190,24 @@ def get_content_info(self):
             desc = get_user_input(settings.content_desc_text, any_input=True)
             self.content_dict['content_description'] = desc
 
+    @staticmethod
+    def _write_to_file_content(file_path, config):
```
bocekm: Why don't you use utils.py/FileHelper/write_to_file()?

preupg/creator/ui_helper.py
```diff
@@ -253,19 +260,37 @@ def _create_group_ini(self):
...
+        srcMajorVersion =
+        dstMajorVersion =
+        """
+        file_name = 'properties.ini'
+        section = 'preupgrade-assistant-modules'
+
+        config = ConfigParser.RawConfigParser()
+        config.add_section(section)
+        config.set(section, 'srcMajorVersion', '')
+        config.set(section, 'dstMajorVersion', '')
```
pirat89: Values should be filled by user during creation of modules in same way as other input is filled.
@AloisMahdal: what do you think about that Lojzo? I am keeping decision on you, but from my side it should be done during creation of modules.

preupg/creator/ui_helper.py
bocekm: Ok, this doesn't save whatever data to a file, it receives a config object and saves it as a file. In this case, the code is fine.

preupg/creator/ui_helper.py
```diff
@@ -195,6 +190,24 @@ def get_content_info(self):
...
+        @param {str} file_path - path of new config file
+        @param {RawConfigParser} config - configuration object with config data
+        @throws {IOError}
+        """
+        try:
+            f = open(file_path, 'wb')
+            config.write(f)
+            f.close()
+        except IOError:
+            print('We have a problem with writing {0} file to disc'.format(
```
bocekm: disc -> disk
OR
remove the "to disc" .. you don't know, what kind of storage is being used :)

preupg/creator/ui_helper.py
```diff
@@ -195,6 +190,24 @@ def get_content_info(self):
...
+    def _write_to_file_content(file_path, config):
+        """
+        Create config file
+
+        @param {str} file_path - path of new config file
+        @param {RawConfigParser} config - configuration object with config data
+        @throws {IOError}
+        """
+        try:
+            f = open(file_path, 'wb')
```
bocekm: Reminding to use "with" statement

preupg/creator/ui_helper.py
```diff
@@ -253,19 +260,37 @@ def _create_group_ini(self):
...
+        file_name = 'properties.ini'
+        section = 'preupgrade-assistant-modules'
+
+        config = ConfigParser.RawConfigParser()
+        config.add_section(section)
+        config.set(section, 'srcMajorVersion', '')
+        config.set(section, 'dstMajorVersion', '')
+
+        self.properties_ini_path = os.path.join(self.get_upgrade_path(), file_name)
+        if os.path.exists(self.properties_ini_path):
```
bocekm: You can move this check to the beginning of the method.

preupg/creator/ui_helper.py
AloisMahdal: @pirat89 I agree.

Just to remind, what the change would mean, p-c-c now works in 2 modes:

 *  found a new set + add new module,
 *  add new module to existing set.

In both cases it creates new module so it has to ask about that.  In second mode, it prepends the module question with for `y/n` confirmation.  Since the src/dst property is property of *the set*, we would want it to ask only in the first mode (set creation).

(The CLI could be improved but for now it's OK.)

AloisMahdal:
Tested  with PR build:  Now preupg-content-generator will succeed, but will create properties.ini with incorrectly named and empty keys:

    # cat AAA/properties.ini 
    [preupgrade-assistant-modules]
    srcmajorversion = 
    dstmajorversion = 

preupg-xccdf-compose will choke on this:

    # preupg-xccdf-compose AAA
    Key srcMajorVersion inside preupgrade-assistant-modules section
    is missing in /root/AAA/properties.ini

MajoDurco:
Yes, that's what @pirat89's comment above was all about.

preupg/creator/ui_helper.py
```diff
@@ -128,7 +126,35 @@ def specify_upgrade_path(self):
         return True
 
     def prepare_content_env(self):
-        self.content_path = os.path.join(self.get_group_name(), self.get_content_name())
+        self.content_path = os.path.join(self.get_group_name(),
+                                         self.get_content_name())
+
+    def properties_ini(self):
```
bocekm: The function name should convey its purpose and should contain a verb (what it does). If it's hard to come up with a good function name then the function is not properly written or decomposed. I would:
- remove this function
- set the `self.properties_ini_path` in the __init__ of UIHelper
- move the `if not self.properties_ini_exists():` check to the `get_properties_ini_versions()` itself

MajoDurco: - You're right I should rename `properties_ini()`
- I'd say `self.properties_ini_path` cannot be placed in `init` method because it will be only set to right value when `UIHelper` takes `upgrade_path` as argument otherwise there is a `specify_upgrade_path` method which ask user for proper path.
- about moving properties file check into `get_properties_ini_versions`:  I didn't want to add additional logic into that function (just ask user 2 questions).

bocekm:
--test--

preupg/creator/ui_helper.py
```diff
@@ -133,7 +124,37 @@ def specify_upgrade_path(self):
...
+
+    def ask_about_properties_ini(self):
+        """ If properties.ini file doesnt exist ask user for OS versions,
+        otherwise don't ask anything
+        """
+        self.properties_ini_path = os.path.join(
+            self.get_upgrade_path(),
+            preupgSettings.properties_ini)
+        if not self.properties_ini_exists():
+            self.get_properties_ini_versions()  # ask user forn  versions
```
bocekm: typo in the comment

preupg/creator/ui_helper.py
```diff
@@ -133,7 +124,37 @@ def specify_upgrade_path(self):
         return True
 
     def prepare_content_env(self):
-        self.content_path = os.path.join(self.get_group_name(), self.get_content_name())
+        self.content_path = os.path.join(self.get_group_name(),
+                                         self.get_content_name())
+
+    def ask_about_properties_ini(self):
```
bocekm: Now you have function _ask_about_properties_ini_ that does not ask user anything directly yet but the function that asks is named _get_properties_ini_versions_.
What about:
- _get_properties_ini_info_ for this one (or _gather_ instead of _get_). Because in case the _properties.ini_ already exists then the function at least gets the path to the file.
- _ask_about_properties_ini_versions_ for the one that asks user.

AloisMahdal:
qa_ack: re-verify with latest version of downstream test suite (`ipu tier pa-*`)

AloisMahdal:
`qa_verified`: The master I mentioned above has been already released, so re-verified with 2.5.0 downstream build+test suite: passed!

Thanks for the great work!

Issue #301, opened 2017-08-29
Required content_description in preupg-content-creator


MajoDurco:
The `preupg-content-creator` question about module description:
(`Would you like to specify the module description? [Y/n]?`) should be removed because
`content_description` is required in `package.ini`. If user doesn't specify it, error is thrown.

MajoDurco:
content_title should be also required.

MajoDurco:
Solved #305.

PR #305, opened 2017-09-01
Content Creator: Make module description and title mandatory

MajoDurco:
Solves #301.


preupg/creator/ui_helper.py
```diff
@@ -356,12 +355,19 @@ def is_valid_string(string):
...
+    def ask_for_string(msg, err_msg):
+        string = get_user_input(msg, any_input=True)
+        if not UIHelper.is_valid_string(string):
             print(err_msg)
             return False
-        return version
+        return string
+
+    @staticmethod
+    def ask_mandatory_string(msg, err_msg):
```
bocekm: Please use this cool new function also within specify_upgrade_path(). Thanks.

AloisMahdal:
qa_ack: manual test should be enough

...although this will actually also heavily affect most of downstream preupgrade-assistant tests (tier "pa-r"), as these use fake modules created by this code.  So once the tests break, are updated and then start working again, we know this has been fixed.

pirat89:
I recommend even replace the *content* by the *module* keyword as the *content* is relic of previous terminology that has been changed time ago. It can be done in separate PR as it will require even change inside modules. But I prefer to change it for planning release yet as we have more changes in modules already accepted.

bocekm:
@pirat89, you mean changing the _content_ to _module_ in the tool name _preupg-content-creator_? Because there's still so many places where _content_ is used and it would be great to do one PR to replace this word everywhere.
BTW, this PR is now waiting for the QA test and then it will be merged.

AloisMahdal:
Now, IIUC, p-c-c asks Y/N whether we want to provide *description*, but title is already mandatory.  So how should this PR affect the title?

bocekm:
@AloisMahdal, the title was mandatory even before, truth to that. But not entering the title by user was handled by a separate function. Here, @MajoDurco used his new function `ask_mandatory_string` that is common for all mandatory inputs, and instead of exiting the tool when the title is not entered, it tells the user that the title is mandatory and asks the user for it again.

AloisMahdal:
Actually it used to accept empty string for both title and description.  I have modified downstream test library to "make up" both strings if test does not provide them and also removed the `y` step.

AloisMahdal:
Verified with downstream test *riskceck* and  preupgrade-assistant-2.4.4-0.201709121548Z.PR305.el6_8.

After mentioned changes to test library, test works fine.

PR #306, opened 2017-09-01
Fix links to files & dirs in Web UI reports

bocekm:
Any link to a file or folder in the Web UI did not work. There were three problems:
- During the upload to the UI, the tarball with the system assessment results was extracted. However, all content in the tarball is under the _preupg_results-\<timestamp\>_ subfolder - the Web UI does not expect this subfolder. Now, just the content that is under this subfolder is extracted.
- Previously, it was possible to display only files (to which there was a link in the Web UI report), not folders. Now, if the user clicks on a link to a folder, the Web UI displays the content of the folder.
- It was possible to view **the original** HTML report through Web UI. Within the report there might be links to files & folders that work correctly in its default location (_/root/preupgrade_) but not in the Web UI. For that, these links need to be updated in the report right after extracting the tarball on the Web UI server.

Resolves [rhbz#1473515](https://bugzilla.redhat.com/show_bug.cgi?id=1473515)

### DIFF
--- a/preupg/ui/report/service.py
+++ b/preupg/ui/report/service.py
@@ -44,16 +44,26 @@ def remove_upload(path):
 
 def extract_tarball(tbpath, target_dir):
     """ return report paths: (xml, html) """
+    def strip_subfolder(tar_content):
+        """The Web UI does not expect the subfolder in the extracted tarball
+        content.
+        """
+        for member in tar_content:
+            path_wo_subfolder = member.path.split(os.sep)[1:]
+            if path_wo_subfolder:
+                member.path = os.path.join(*member.path.split(os.sep)[1:])
+                yield member
+
     tar = tarfile.open(tbpath)
-
     tar_content = tar.getmembers()
-
-    xml = filter_files_by_ext(tar_content, '.xml', 'Missing XML report in tarball.')
-    html = filter_files_by_ext(tar_content, '.html', 'Missing HTML report in tarball.')
-
-    tar.extractall(path=target_dir)
-
+    tar.extractall(target_dir, strip_subfolder(tar_content))
     tar.close()
+
+    xml = filter_files_by_ext(tar_content, '.xml',
+                              'Missing XML report in tarball.')
+    html = filter_files_by_ext(tar_content, '.html',
+                               'Missing HTML report in tarball.')
+
     xml_path = os.path.join(target_dir, xml.name)
     html_path = os.path.join(target_dir, html.name)
     return xml_path, html_path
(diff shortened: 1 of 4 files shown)


preupg/ui/report/service.py
```diff
@@ -44,16 +44,26 @@ def remove_upload(path):
 
 def extract_tarball(tbpath, target_dir):
     """ return report paths: (xml, html) """
-    tar = tarfile.open(tbpath)
+    def strip_subfolder(tar_content):
+        """The Web UI does not expect the subfolder in the extracted tarball
+        content.
+        """
+        for member in tar_content:
+            path_wo_subfolder = member.path.split(os.sep)[1:]
```
MajoDurco: Seems like typo `path_to_subfolder`.

preupg/ui/report/tests.py
```diff
@@ -82,7 +82,7 @@ def test_stringify_children(self):
 #         tarball_path = a.scan_system()
 #
 #         xml_path, html_path = extract_tarball(tarball_path, self.temp_dir)
-#         self.assertTrue(len(parse_report(xml_path)) > 0)
```
MajoDurco: Does this test suite have any purpose here when it's commented or can it be removed?

preupg/ui/report/service.py
bocekm: It's meant is without :) I'll change it.

preupg/ui/report/tests.py
bocekm: Those tests were run very long time ago and I'm not even sure how to run them. This should probably be reported as a new issue to be resolved.

AloisMahdal:
qa_ack: Follow steps in rhbz#1473515.

pirat89:
Just for info, I looked at some another default outputs on our servers and there is used this format, that looks better (from my point of view) then the used here.
```
[DIR] dirname/
[   ] filename
```

Maybe it would be modified yet, the output will be more readable. I recommend use the "underscore" inside brackets instead of space, so the content will be aligned correctly. See:

[DIR] dirname/
[___] filename

bocekm:
@pirat89, I haven't seen the format you're proposing yet. This is in fact quite subjective matter, but I'm not in favor of it personally. Actually, even with the underscores it's not aligned correctly. Let's not fall under the spell of [bike-shed effect](https://en.wikipedia.org/wiki/Bike_shed_effect).

pirat89:
OK. It was not requirements and agree that there are more important things.

bocekm:
Thanks @MajoDurco for the review. Now the PR awaits QA test before I go ahead and merge it.

bocekm:
@AloisMahdal, PR rebased, so it includes your PR #314.

AloisMahdal:
Tested with preupgrade-assistant-2.4.4-0.201709191850Z.PR306.el6_8 and preupgrade-assistant-el6toel7-0.6.71-1.201709181350Z.el6toel7.el6_8.

Looking at xccdf_preupg_rule_backup_NoverifyConfigs_check, links to files as well as directories work fine.

Issue #280, opened 2017-07-24
Do not require common/release_version in a module set


bocekm:
Currently when running Preupgrade Assistant with a custom user-created module set, the tool requires a _release_version_ file within the _common_ subdirectory of the module set directory. @pirat89 Does Preupgrade Assistant really need this file for run? If yes, we should clarify that in a module set-writing tutorial.

bocekm:
The mentioned file is used only by the function _get_system_versions_ in [crossarch_migration_support_check.patch](https://github.com/upgrades-migrations/preupgrade-assistant/blob/2.4.4/packaging/sources/crossarch_migration_support_check.patch)

Because of the implementation of PR https://github.com/upgrades-migrations/preupgrade-assistant/pull/233 (each module set now needs _properties.ini_ with the source and target OS version), this piece of code can be updated to read the _properties.ini_ instead of the _common/release_version_.

Issue #279, opened 2017-07-24
Update crossarch_migration_support_check.patch to allow custom contents


bocekm:
Currently the [crossarch_migration_support_check.patch](https://github.com/upgrades-migrations/preupgrade-assistant/blob/master/packaging/sources/crossarch_migration_support_check.patch) requires proprietary data from the package _preupgrade-assistant-el6toel7_ (because of `/root/preupgrade/RHEL6_7/common/release_version` file) and _preupgrade-assistant-el6toel7-data_ (because of `/root/preupgrade/RHEL6_7/common/<src_arch>-<dst_arch>` folder). That shouldn't be happening because Preupgrade Assistant should be able to run with a custom user-created module set, i.e. without the proprietary data.

AloisMahdal:
This is obsoleted by #314, which replaces the currently used method with another one,  removing the patch completely (ie. moving the relevant code to module set, hence [modules' #12][1]).

  [1]: https://github.com/upgrades-migrations/preupgrade-assistant-modules/pull/12

PR #314, opened 2017-09-13
Module set init script

AloisMahdal:
This PR contains two teps:

 *  Implement "module set init" feature, where a module set may contain `init` executable,
    and if it does, p-a will run it before module set and fail if errors are found.

 *  Move current code of `prep_links()` (and associated patch that modifies it) from
     common.sh to module set; hence the other PR in preupgrade-assistant-modules.


AloisMahdal:
Note that this PR fixes #280 and obsoletes #279.

AloisMahdal:
Thanks for feedback.  The new error handling is as follows:

 *  if file does not exist, nothing happens.

 *  if it does, preupg will always attempt to launch it but any `OSError` is caught and
    re-raised as `ModuleSetFormatError` (a new exception class I've added for purposes
    of malformed module sets).

This should cover more cases than just file is not being executable; also cases when file is a directory or (I think) malformed shebang line.  And yes, preupg crashes; let's err on the safe side.

I also made few minor cleanups within the code added by first commit:

 *  I removed the `return True` at the end; the value is not consumed by anything. (Frankly
     I don't know why I even added it in the first place).

 *  I fixed typo in the `ModuleSetInitError` docstring (`s/file/script/`).

 *  I fixed formatting of `scenario` assignment (spaces around `=` were missing).

AloisMahdal:
Oops, made some errors in the last commit, please wait for update.

AloisMahdal:
OK, so errors have been fixed:

 *  fixed exception reference (`exception.ModuleSetFormatError`, not just
     `ModuleSetFormatError`)

 *  Improved `ModuleSetFormatError` message to include path and string from OSError.

AloisMahdal:
I have also tested it against following cases:

 *  no init,
 *  init is a non-executable file,
 *  init is a directory,
 *  init is executable but empty,
 *  init is executable but with none or bad shebang,
 *  init is executable and passed,
 *  init is executable and failed (stderr or exit code 1).

Also I've re-run result-files test from downstream suite.

pirat89:
It works correctly know. Just in case that the `init` is directory, the error message looks little weird and something like info that the script is expected instead of directory would be better, but it is just littleness and in addition it is improbable situation.

The modules are already prepared for this change so it can be merged.

AloisMahdal:
Thanks, updated:

 *  `init_env` now holds just the extra variables. `os.environ` is copied and updated when
    passed to `Popen ` so that we can keep extra variables in `init_env`

 *  `init_env` forms most of the body of the debug message, which is displayed only if
     init is about to be launched.

 *  if init script is not found, nothing is logged.

AloisMahdal:
Sorry, I did not realize `dict.update()` updates *in-place* and returns error.

I have pushed new version with fixed dict logic.

AloisMahdal:
qa_verified: already done earlier; adding labels in hindsight.

PR #316, opened 2017-09-13
Remove dead functions and classes

AloisMahdal:
A spontaneous followup to #313.

These functions and class are nor part of public API nor are they
mentioned anywhere else in the project (checked using `grep -r`.)

### DIFF
--- a/preupg/cli.py
+++ b/preupg/cli.py
@@ -26,17 +26,6 @@ def upload_callback(option, dummy_opt_str, dummy_value, parser):
                                            " upload option.")
 
 
-def optional_rh_arg(arg_default):
-    def func(option, opt_str, value, parser):
-        if parser.rargs and not parser.rargs[0].startswith('-'):
-            val = parser.rargs[0]
-            parser.rargs.pop(0)
-        else:
-            val = arg_default
-        setattr(parser.values, option.dest, val)
-    return func
-
-
 class CLI(object):
 
     """Class for processing data from commandline"""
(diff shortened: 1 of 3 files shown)


preupg/utils.py
```diff
@@ -383,12 +383,6 @@ def run_subprocess(cmd, output=None, print_output=False, shell=False, function=N
         return sp.returncode
 
 
-class PreupgHelper(object):
-    @staticmethod
-    def get_prefix():
-        return settings.prefix
-
-
```
pirat89: This is just relic of *preupgrade*/*premigrate* history.

preupg/cli.py
```diff
@@ -26,17 +26,6 @@ def upload_callback(option, dummy_opt_str, dummy_value, parser):
...
-    def func(option, opt_str, value, parser):
-        if parser.rargs and not parser.rargs[0].startswith('-'):
-            val = parser.rargs[0]
-            parser.rargs.pop(0)
-        else:
-            val = arg_default
-        setattr(parser.values, option.dest, val)
-    return func
-
-
```
pirat89: It was used originaly for the `--non-rh-signed` option that has been removed time ago.

preupg/logger.py
```diff
@@ -75,19 +75,3 @@ def log_message(message, new_line=True, level=logging.INFO):
...
-        sys.stdout.flush()
-        # This is used in case that we do not want to print the new line
-        if new_line:
-            sys.stdout.write("\n")
-            sys.stdout.flush()
-    else:
-        endline = "\n" if new_line else ""
-        print(message, end=endline, file=sys.stdout, flush=True)
-
-    logger_report.log(level, message)
```
pirat89: This functions is obsoleted since the 43c1883 commit.

PR #319, opened 2017-09-22
Support older versions of the OpenSCAP

pirat89:
Old OpenSCAP doesn't support new complex report style that is used
by default by the Preupgrade Assistant tool. Enforce the simple
style of report (like with --old-report-style) when version of
installed OpenSCAP is lower then 1.2.7. In case of enforced fallback
to the simple style of report log message to inform the user.

Additionaly change requirements to openscap* >= 0:1.0.10

### DIFF
--- a/packaging/preupgrade-assistant.spec
+++ b/packaging/preupgrade-assistant.spec
@@ -48,9 +48,9 @@ Requires:       sed findutils bash
 Requires:       rpm-python
 Requires:       redhat-release
 Requires:       yum-utils
-Requires:       openscap%{?_isa} >= 0:1.2.8-1
-Requires:       openscap-engine-sce%{?_isa} >= 0:1.2.8-1
-Requires:       openscap-utils%{?_isa} >= 0:1.2.8-1
+Requires:       openscap%{?_isa} >= 0:1.0.10
+Requires:       openscap-engine-sce%{?_isa} >= 0:1.0.10
+Requires:       openscap-utils%{?_isa} >= 0:1.0.10
 Requires:       pykickstart
 Requires:       python-six
 Conflicts:      %{name}-tools < 2.1.0-1
--- a/preupg/application.py
+++ b/preupg/application.py
@@ -104,6 +104,8 @@ class Application(object):
         self.tar_ball_name = None
         self.third_party = ""
         self.assessment_dir = None
+        self._set_old_report_style()
+
 
     def _add_report_log_file(self):
         """
@@ -142,6 +144,24 @@ class Application(object):
                            settings.preupg_log)
         else:
             self.debug_log_file = settings.preupg_log
+
+    def _set_old_report_style(self):
+        """
+        Choose which HTML report style should be used.
+
+        Set self.old_report_style to True when it is required from commandline
+        or older version of OpenSCAP has been detected.
+        """
+        if self.conf.old_report_style:
+            self.old_report_style = True
+        elif OpenSCAPHelper.is_oscap_equal_or_greater(1, 2, 7) is False:
+            # in case that OpenSCAP version is lower then 1.2.7, fallback
+            # to the old (simple) report style.
+            log_message("Generating simply styled report due to the "
+                        + "limitations of the installed OpenSCAP")
+            self.old_report_style = True
+        else:
+            self.old_report_style = False
 
     def get_third_party_dir(self, assessment):
         """
@@ -355,7 +375,7 @@ class Application(object):
         self.report_parser.reload_xml(self.openscap_helper.get_default_xml_result_path())
         self.report_parser.update_check_description()
         xml_report = self.openscap_helper.get_default_xml_result_path()
-        if self.conf.old_report_style:
+        if self.old_report_style:
             ReportParser.write_xccdf_version(xml_report, direction=True)
 
     def generate_html_or_text(self):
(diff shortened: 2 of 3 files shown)


packaging/preupgrade-assistant.spec
```diff
@@ -44,9 +44,9 @@ Requires:       sed findutils bash
 Requires:       rpm-python
 Requires:       redhat-release
 Requires:       yum-utils
-Requires:       openscap%{?_isa} >= 0:1.2.8-1
-Requires:       openscap-engine-sce%{?_isa} >= 0:1.2.8-1
-Requires:       openscap-utils%{?_isa} >= 0:1.2.8-1
+Requires:       openscap%{?_isa} >= 0:1.0.10
+Requires:       openscap-engine-sce%{?_isa} >= 0:1.0.10
```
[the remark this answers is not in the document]
bocekm: I've tried to run `preupg` on:
- RHEL 6.5 - no go, there's no easy workaround for the missing openscap-engine-sce
- RHEL 6.6 - I did get a traceback:
   ```
  File "/usr/lib/python2.6/site-packages/preupg/utils.py", line 537, in config_has_section
    config = configparser.RawConfigParser(allow_no_value=True)
  TypeError: __init__() got an unexpected keyword argument 'allow_no_value'
   ```
   So I removed the `allow_no_value=True` and then the assessment worked and the old (simple) report was generated. This needs to get fixed.

I suggest that we will no longer state anywhere that the Preupgrade Assistant can run on RHEL 6.5.
a) We can probably have the Lab updated:
  - https://access.redhat.com/labs/rhelupgradehelper/
    Use product Red Hat Customer Portal Labs and component RHEL Upgrade Helper

b) No KB article state that the Preupgrade Assistant can run on RHEL 6.5, so there's no update to any of them needed.
c) Update the Requires in spec to:
   ```
   openscap%{?_isa} >= 0:1.0.8
   openscap-engine-sce%{?_isa} >= 0:1.0.8
   openscap-utils%{?_isa} >= 0:1.0.8
   ```

preupg/application.py
```diff
@@ -346,12 +346,24 @@ def clean_scan(self):
     def prepare_for_generation(self):
         """Function prepares the XML file for conversion to HTML format"""
         for report in self._get_reports():
+            # at first, check which report style should be used.
```
bocekm: Do not use "at first" if you don't use "but" in the same sentence. Use "First" instead.

preupg/utils.py
```diff
@@ -762,6 +762,41 @@ def update_variables(self, result_dir, result_name, xml_result_name, html_result
...
+            if oscap_version["major"] >= 0:
+                return
+            # this should happen only in case of first line
+            rmatch = re.match(r".*\s([0-9]+)\.([0-9]+)\.([0-9]+)\s$", oscap_msg)
+            if rmatch is None:
+                return
+            oscap_version["major"] = int(rmatch.group(1))
+            oscap_version["minor"] = int(rmatch.group(2))
+            oscap_version["patch"] = int(rmatch.group(3))
+        ProcessHelper.run_subprocess(["/usr/bin/oscap", "version"],
```
bocekm: Use `openscap_binary` from settings.py

preupg/utils.py
```diff
@@ -762,6 +762,41 @@ def update_variables(self, result_dir, result_name, xml_result_name, html_result
...
+        Get version of the OpenSCAP.
+
+        Return version of the openscap as dict with the 'major', 'minor' and
+        'patch' keys.
+
+        Raise OSError when the oscap utility is not installed.
+        """
+        oscap_version = {"major" : -1, "minor" :  -1, "patch" : -1}
+        def _process_oscap_output(oscap_msg):
+            if oscap_version["major"] >= 0:
```
bocekm: How can ever be `oscap_version["major"] >= 0` here?

preupg/utils.py
```diff
@@ -762,6 +762,41 @@ def update_variables(self, result_dir, result_name, xml_result_name, html_result
...
+                return
+            # this should happen only in case of first line
```
bocekm: This comments seems a bit confusing to me? Did you want to say that the version of OpenSCAP appears on the first line? Please reword the sentence.

preupg/application.py
pirat89: Thanks :-)

preupg/utils.py
pirat89: Exactly :-)

preupg/utils.py
pirat89: Added comment. The callback function is called for each line of stdout separately.

packaging/preupgrade-assistant.spec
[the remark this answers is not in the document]
pirat89: I prefer to move evertyhing around RHEL-6.6 to separate PR because the parser returns error during parsing becuuse of lines where only values are used. It's interesting as it maby points to another error in PA when the config has been "sucessfully" parsed on RHEL-6.6. Probably some catch of the exception?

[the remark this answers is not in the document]
bocekm: Ok. I created an issue for that: https://github.com/upgrades-migrations/preupgrade-assistant/issues/322.

bocekm:
@AloisMahdal, please provide a qa_ack or qa_nak.

AloisMahdal:
qa_ack: Manual test only.

Just run an assessment (and 'ast' test with mock module should be fine) with 6.7 and 6.6 and review that results.html is generated correctly.

bocekm:
@AloisMahdal, just with RHEL 6.7. For RHEL 6.6 there's an issue created to support it: #322

pirat89:
Rebased with current master and modified code.

AloisMahdal:
Tested with preupgrade-assistant-2.4.4-0.201710161905Z.PR319.el6_8.noarch on RHEL-6.7.  results.html was generated and looks fine.

PR #327, opened 2017-10-16
Bump to version 2.5.0

bocekm:
**New:**
- Name of a module set directory can now be arbitrary
- Module set directory now requires properties.ini file (see https://github.com/upgrades-migrations/preupgrade-assistant/blob/master/doc/module_writing_tutorial/README.md for its content format)
- Module set can now contain an executable `init` script which is executed after the "common" files-gathering stage and before execution of the modules
- Possible to skip building UI now

**Changed:**
- READMEs moved from `/usr/share/preupgrade` to `/usr/share/doc`
- Removed CPE platform from the result report
- Removed obsolete `RESULT_FAILED` and `check_root()` from bash API
- Removed obsolete `exit_failed()` from python API
- Removed deprecated `preupg-create-group-xml` utility
- It's not allowed anymore to have `solution` and `check_script` keys in module.ini - the solution file and check script need to named _solution.txt_ and _check_

**Fixed:**
- `<module set directory>/common/release_version` file and `<module set directory>/common/<src_arch>-<dst_arch>` directory are now not required
- Before the _stdout_ and _stderr_ was not shown for the _pass_ and _fixed_ results in the simple style report (`--old-report-style`). Now they are shown no matter what the result of a module is.
- `preupg-content-creator`: Not entering module's _description_ or _title_ could break processing of a module set. They are mandatory to input now.
- Module set `all-xccdf.xml` is not being corrupted now in case of using the `--contents` option
- Fixed links to files and directories in Web UI reports
- The Preupgrade Assistant can run again on RHEL 6.7. The previous release could run RHEL 6.8+.


pirat89:
Agree with Lojza. pr #311  should be included.

PR #328, opened 2017-10-17
Creator: do not insert 'solution:' to module.ini

bocekm:
preupg-content-creator was adding `solution` key to the module INI.
The module's solution file has to be named _solution.txt_ and the solution key is not supposed to be in the module INI anymore.
Related: https://github.com/upgrades-migrations/preupgrade-assistant/pull/311

### DIFF
--- a/preupg/creator/ui_helper.py
+++ b/preupg/creator/ui_helper.py
@@ -10,7 +10,7 @@ import shutil
 import sys
 
 from distutils.util import strtobool
-from preupg.utils import FileHelper, ModuleSetUtils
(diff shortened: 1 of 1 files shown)


AloisMahdal:
Updated mentioned test and ran with preupgrade-assistant-2.4.4-0.201710171159Z.PR328.el6_8.noarch; behavior is as expected.

bocekm:
@AloisMahdal, @pirat89, @jmazanek, the PR is verified but I need a code review. Would anyone of you, please, do that?

Issue #320, opened 2017-10-01
riskcheck will not run without upgrade path at /usr


AloisMahdal:
Importing rhbz#1372090.  Verified with preupgrade-assistant-2.4.4-0.201709220805Z.PR311.el6_8.noarch; it's still affected.

---

Description of problem
----------------------

It's possible to generate assessment without upgrade path installed at /usr/share, but it's not possible to use --riskcheck to verify this assessment: preupg will complain that upgrade path is not installed and exit with status 20.

Version-Release number of selected component
--------------------------------------------

preupgrade-assistant-2.1.9-1.el6

How reproducible
----------------

Always

Steps to Reproduce
------------------

 1. Create a custom upgrade path
 2. Run preupg with this path (--contents parameter)
 3. Run preupg --riskcheck --verbose

Actual results
--------------

Error is printed:

    There were no contents found in directory /usr/share/preupgrade. If you would like to use this tool, you have to install some

and exit status is 20.

Expected results
----------------

No error

Additional info
---------------

Note that --riskcheck (and other modes) documentation does not mention way to specify *which* result to check, so there's an underlying assumption that it's always the last one.

bocekm:
--riskcheck functionality has been overhauled in https://github.com/upgrades-migrations/preupgrade-assistant/pull/218. I'll check whether it fixes your issue.

bocekm:
Yay, still the same issue :(

bocekm:
However `preupg --riskcheck -c /tmp/RHEL6_7/all-xccdf.xml --verbose` works. @AloisMahdal, so you think it's insufficient?

PR #329, opened 2017-10-18
Allow certain operations without module set present

bocekm:
Before, when _--upload_, _--riskcheck_ or --cleanup was used with no module set in _/usr/share/preupgrade_, the Preupgrade Assistant failed, as it was trying to find any module set. However no module set is needed for these two options, they are supposed to work solely with the latest assessment result.

Resolves https://github.com/upgrades-migrations/preupgrade-assistant/issues/320

### DIFF
--- a/preupg/application.py
+++ b/preupg/application.py
@@ -696,6 +696,38 @@ class Application(object):
                 log_message("%s" % dir_name)
             return 0
 
+        if self.conf.riskcheck:
+            result_xml_path = os.path.join(settings.assessment_results_dir,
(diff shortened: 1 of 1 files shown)


AloisMahdal:
`qa_ack`: The *--riskcheck* part already causes downstream test preupg/riskcheck to fail, so re-run that test.  If the test fails, rest is enough to check manually.

AloisMahdal:
`qa_verified`: verified with 2.5.0-0.201710181910Z.PR329.el6_8 by running test suite; all passed (a rare sight!)

PR #340, opened 2017-11-01
Use the path to the copied module set directory instead of dirname

pirat89:
The "placeholer" was not replaced, because XmlManager expects path
to the copied module set (/root/preupgrade/RHEL6_7) instead
of dirname (RHEL6_7).


AloisMahdal:
`qa_ack`: downstream test suite now covers this as part of preupg/result-files. (It was covered  before but the placeholder string has changed since so I've updated the test.)

AloisMahdal:
Tested with `preupgrade-assistant-2.5.0-0.201711012235Z.PR340.el6_8`; solution placeholers are not present and texts are properly shown in HTML (verified, I promise!).

PR #342, opened 2017-11-02
Rename variables to include 'path' instead of 'dir'

bocekm:
These variables actually hold the whole path and not just a dirname. It has already caused confusion and error - see https://github.com/upgrades-migrations/preupgrade-assistant/pull/340.


bocekm:
--test--

PR #348, opened 2018-01-09
Fix running 3rd party modules

bocekm:
Enable processing of third party modules. Original solution has been completely dropped (with additional dead code) because it was too messy and buggy. Current solution regenerates original XCCDF compose under `assessment_results_dir` from the used module-set. Limitation of this solution is processing of the third party modules in group with the others instead of previous idea to process them after the *main* set of modules is processed.

~~TODO~~ (Done):
  - handle situation when src and dst paths for XCCDFCompose are same. Currently the dst path is alwasy removed when exist  and this could lead into lost of original modules. This is not possible now, but it would in future. It will be better to handle it now to prevent such situation in future. Possible solution
    -  is to ignore given dst path and behave in the same way as dst path is not set
    - or raise an exception (preferred one) (**chosen one**)

### DIFF
--- a/preupg/application.py
+++ b/preupg/application.py
@@ -399,7 +399,16 @@ class Application(object):
         return module_set_dirname
 
     def prepare_scan_system(self):
-        """Function cleans previous scan and creates relevant directories"""
+        """
+        Prepare system for scan.
+
+        In case the result of previous assessment is detected, remove it
+        and prepare new directory structure that will be used for next scan.
+        Additionaly process scripts generating common data about the current
+        system, generate final XCCDF compose from the original set of modules
+        identified for run and in the end process initial script of modules to
+        complete expected environment for the scan.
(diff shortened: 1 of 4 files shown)


bocekm:
It looks good now. Thanks, @pirat89, for help.

bocekm:
@AloisMahdal, could you please do a verification? Thanks.

AloisMahdal:
`qa_ack+`: Add "naked" module into *3rdparty* directory and verify if it runs.  (Already covered downstream in *result-files*.)

AloisMahdal:
(Adding `on_qe` per agreement with Michal)

AloisMahdal:
`verified`: ran with *preupgrade-assistant-2.5.1-0.201801161720Z.PR348.el6_8.noarch*.  Test passes.

PR #350, opened 2018-02-09
The list of files should contain only executable files, not symlinks

shaded-enmity:
Add `-type f` argument for `find` so that it does not print symlinks
that might point to non-local file systems


bocekm:
@AloisMahdal, do you feel it's worth verifying from your side? If not, I'd merge.

bocekm:
@AloisMahdal agrees with not verifying it.

PR #351, opened 2018-02-09
Remove patch that adds commands to scripts.txt

bocekm:
The commands that are added by the patch are RHEL-specific however the Preupgrade Assistant API for modules requires them so the whole tool can be considered as RHEL-specific. The attempts to have the tool for Fedora too have been abandoned for a long time.


AloisMahdal:
Can you describe what these lines do in terms of preupgm API?   Then I could add some basic coverage in test suite; otherwise I'm not sure what to test here.

pirat89:
Not sure whether we really need to cover this by tests. Basically, now it should be ok just verify, that installed file hasn't been changed before and after change. In that case everything is same as it was before.

bocekm:
The scripts.txt contains commands that generate files. The added five commands from the patch generate five files, which are used in API:
```
VALUE_CHKCONFIG=$PREUPGRADE_CACHE/chkconfig.log
VALUE_RPMTRACKEDFILES=$PREUPGRADE_CACHE/rpmtrackedfiles.log
VALUE_ALLMYFILES=$PREUPGRADE_CACHE/allmyfiles.log
VALUE_EXECUTABLES=$PREUPGRADE_CACHE/executable.log
VALUE_RPM_RHSIGNED=$PREUPGRADE_CACHE/rpm_rhsigned.log
```
The constants _VALUE__ can be used (and are used) by RHEL 6->7 upgrade modules.

bocekm:
@AloisMahdal, a test may be created to check whether these files are being created during the PA execution.

AloisMahdal:
`qa_ack`: add test with a mock module which only tests for existence of these files.

Thanks!

tcerna:
Hi, I checked source packages of preupgrade-assistant and patch was removed and all changes it made was moved to data/preassessment/scripts.txt. All necessary logs mentioned above was created.

PR #352, opened 2018-02-11
Build XCCDF only from module.ini and group.ini files

### LINKED ISSUE #310: Multilpe ini files inside module
When PA generates `group.xml` file for each module it ussualy collects data only from `module.ini`. Problem is when in module are multiple ini files, then it generates wrong xml (with multiple roots elements).

Error looks like:
```
ExpatError: Encountered a parse error in file /tmp/preupgradeSOPbRF/FOOBAR6_7/failed/group.xml.
Details: junk after document element: line 51, column 0
```
From:
https://github.com/upgrades-migrations/preupgrade-assistant/blob/7290804ad755099bcdc3d5b9a4fd13433fa19712/preupg/xmlgen/oscap_group_xml.py#L77

pirat89:
Previously building of module set has been corrupted when directory
    with the module contained more than one file with the ".ini"
    extentions. Build of the module set shouldn't be affected by various
    INI files, that are not related to the module set structure.
    
    So the scan is now read only group.ini and module.ini files whose
    names are reserved for PA purposes. In addition we expect only one
    of those files in each directory, so the ModuleSetFormatError
    exception is raised when both are detected in the same directory.
   
    Additionaly: 
      - removed unused import of ModuleSetUtils
      - modified tests
    
    Resolves #310

### DIFF
--- a/preupg/settings.py
+++ b/preupg/settings.py
@@ -46,6 +46,7 @@ log_dir = "/var/log/preupgrade"
 # file with module set meta info
 properties_ini = "properties.ini"
 module_ini = "module.ini"
+group_ini = "group.ini"
 
 solution_txt = "solution.txt"
 check_script = "check"
--- a/preupg/xmlgen/oscap_group_xml.py
+++ b/preupg/xmlgen/oscap_group_xml.py
@@ -14,12 +14,12 @@ except ImportError:
     import ConfigParser as configparser
 
 from preupg.xmlgen.xml_utils import XmlUtils
-from preupg.utils import FileHelper, ModuleSetUtils
+from preupg.utils import FileHelper
 try:
     from xml.etree import ElementTree
 except ImportError:
     from elementtree import ElementTree
-from preupg import settings
+from preupg import settings, exception
 
 try:
     from xml.etree.ElementTree import ParseError
@@ -56,9 +56,19 @@ class OscapGroupXml(object):
         self.loded dict:
         self.loaded[ini_file_path] = {option1: value, option2: value, ...}
         """
-        for dir_name in os.listdir(self.dirname):
-            if dir_name.endswith(".ini"):
-                self.lists.append(os.path.join(self.dirname, dir_name))
+        found = 0
+        for file_name in os.listdir(self.dirname):
+            if file_name not in [settings.module_ini, settings.group_ini]:
+                continue
+            if found == 0:
+                self.lists.append(os.path.join(self.dirname, file_name))
+                found = 1
+            else:
+                raise exception.ModuleSetFormatError(
+                        "Cannot prepare XCCDF for OpenSCAP",
+                        "group.ini and module.ini are in the same directory %s"
+                        % self.dirname
+                        )
         for file_name in self.lists:
             if FileHelper.check_file(file_name, "r") is False:
                 continue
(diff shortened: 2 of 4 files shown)


pirat89:
This is more like nice to have now, but I have good mood on Sunday.

AloisMahdal:
CASES
=====


Legend
------

 *  INPUT: tree - contents of directories above; each character is for one
    level, `p` is for properties.ini, `g` is for group.ini and `.` is for
    empty

    E.g. `pg.g` means that the tree structure is something like:

        RHEL6_7
         |- properties.ini
         '- foo
             |- group.ini
             '- bar
                 '- baz
                     |- group.ini
                     '- testdir

    where the fields `gini`, `mini` and `fini` relate to contents of the `testdir`
    directory.

 *  INPUT: gini - group.ini present?
 *  INPUT: mini - module.ini present?
 *  INPUT: fini - foo.ini present?

 *  BEHAVIOR: result - effect in result.xml

 *  `~` - does not matter


Table
-----

    INPUT                      || BEHAVIOR
    ------.------.------.------||----------------------------
     tree | gini | mini | fini || result
    ======|======|======|======||============================
     p    |   no |   no |   no || none (no relevant file)
     p    |   no |   no |  yes || none (no relevant file)
     p    |   no |  yes |   no || none (no group)
     p    |   no |  yes |  yes || none (no group)
     p    |  yes |   no |   no || none (empty group)
     p    |  yes |   no |  yes || none (empty group)
     p    |  yes |  yes |   no || error
     p    |  yes |  yes |  yes || error
    ------|------|------|------||----------------------------
     pg   |   no |   no |   no || none (empty group)
     pg   |   no |   no |  yes || none (empty group)
     pg   |   no |  yes |   no || normal result
     pg   |   no |  yes |  yes || normal result
     pg   |  yes |   no |   no || none (empty group)
     pg   |  yes |   no |  yes || none (empty group)
     pg   |  yes |  yes |   no || error
     pg   |  yes |  yes |  yes || error
    ------|------|------|------||----------------------------
     p.g  |    ~ |    ~ |    ~ || none (chain interrupted)
    ------|------|------|------||----------------------------

AloisMahdal:
`qa_ack`: add and run test for module discovery according to case table above.

bocekm:
@pirat89, I did some code cleanup, please take a look. The code now looks for just one ini file (either module.ini or group.ini) instead of multiple. If there's no one of these two in a folder, it's a failure.

pirat89:
> If there's no one of these two in a folder, it's a failure.
In my testing it is not true. But as we agreed, this could be probably removed at all.

bocekm:
@pirat89 has found issues with the code. These are my notes that may not make much sense to other:
It does not yell when there's no INI in a folder. Which should not actually - there may be folder under module storing just some data. But the code tells to die when there's no INI in any folder (recursive search through all RHEL6_7 module set folder).

pirat89:
Correct. If I hadn't known about the issues, I would not be able to decrypt notes above :-D

AloisMahdal:
Cleaning up qa_ack from last milestone.

AloisMahdal:
`qa_ack`: same as above: add and run test for module discovery according to case table above.

I think I have test in making somewhere; couldn't find it right now, though.

bocekm:
@AloisMahdal ready to be verified.

bocekm:
That's strange, our Jenkins jobs for CD has started by themself and failed because they were started properly. I'm gonna push force causing another rebuild.

pirat89:
-- test --

pirat89:
--test--

pirat89:
Just rebased to run the build. without any other changes.

pirat89:
--test--

PR #353, opened 2018-02-15
Fix reading VALUE_EXECUTABLES file with non-UTF-8 chars

bocekm:
The Python API function `check_rpm_to` was calling `get_file_content()` (with UTF-8 decoding flag being True) that **had no purpose there** and it was causing tracebacks for users when the _executable.log_ (`VALUE_EXECUTABLES`) file contained some non-UTF-8 characters.

```
Traceback (most recent call last):
  File "/root/preupgrade/RHEL6_7/system/RemovableMedia/check", line 24, in &lt;module&gt;
    check_rpm_to (check_rpm="",check_bin="lsblk")
  File "/usr/lib/python2.6/site-packages/preupg/script_api.py", line 526, in check_rpm_to
    lines = FileHelper.get_file_content(VALUE_EXECUTABLES, "rb", True)
  File "/usr/lib/python2.6/site-packages/preupg/utils.py", line 186, in get_file_content
    data = f.read() if not method else f.readlines()
  File "/usr/lib64/python2.6/codecs.py", line 679, in readlines
    return self.reader.readlines(sizehint)
  File "/usr/lib64/python2.6/codecs.py", line 588, in readlines
    data = self.read()
  File "/usr/lib64/python2.6/codecs.py", line 477, in read
    newchars, decodedbytes = self.decode(data, self.errors)
UnicodeDecodeError: 'utf8' codec can't decode byte 0xa2 in position 132483621: invalid start byte
```

Related cleanup:
The method for opening a file without decoding `codecs.open()` is not needed, built-in `open()` is sufficient.

### DIFF
--- a/preupg/script_api.py
+++ b/preupg/script_api.py
@@ -513,7 +513,6 @@ def check_rpm_to(check_rpm="", check_bin=""):
 
     if check_bin != "":
         binaries = check_bin.split(',')
-        lines = FileHelper.get_file_content(VALUE_EXECUTABLES, "rb", True)
         for binary in binaries:
             cmd = "which %s" % binary
             if ProcessHelper.run_subprocess(cmd, print_output=False, shell=True) != 0:
--- a/preupg/utils.py
+++ b/preupg/utils.py
@@ -183,7 +183,7 @@ class FileHelper(object):
         if decode_flag:
             f = codecs.open(full_path, perms, settings.defenc)
         else:
-            f = codecs.open(full_path, perms)
+            f = open(full_path, perms)
         try:
             data = f.read() if not method else f.readlines()
         finally:


AloisMahdal:
What do you mean by "improperly encoded characters"? Or "improper characters"?

Files just contains bytes, without knowledge of encoding, there is no way of knowing if the bytes should even be translated to characters, let alone how.  And with most 8bit encodings, any byte just does have meaning, so  there's no way of discerning between incorrect characters and correct ones.  UTF-8 is a notable exception (well it's not really 8-bit encoding), but character sequences invalid in UTF-8 can be perfectly valid in other encodings, so there's no way of knowing just from the file itself.

Can you assume that whoever produced the executable.log *meant* it to be UTF-8?  If yes (it's our code), then there's nothing to test except maybe on unit-level. If no, then you have to know the encoding (from other source than the file itself) prior to opening the file.

bocekm:
@AloisMahdal, you're correct, improperly encoded was not a good wording. I don't know what encoding has been used when saving the executable.log file. Certainly the characters were neither ASCII nor UTF-8 encoded. The issue was that in our code we were opening the executable.log file with UTF-8 encoding but we didn't use the opened file for anything.

BTW, out of curiosity, I used `chardet` to try to detect encoding and it said Turkish. When I opened the file with the encoding, it was still garbage: `/opt/../CLI/¢À¿àçƒqò¶z—¶Ïz�‡ê“Â*Ë·U¨%·£` so it more like a corrupted data.
```
In [5]: chardet.detect(open("/tmp/executable_full.log", "rb").read())
Out[5]: 
{'confidence': 0.5208752096155429,
 'encoding': 'Windows-1254',
 'language': 'Turkish'}
```
But because in the python API we open all the common generated files with UTF-8 encoding, I'll try to change system encoding to some non-UTF-8 to see what will happen.

bocekm:
This PR does not fix the problem of failing modules when non-UTF-8 locale is used so I created issue https://github.com/upgrades-migrations/preupgrade-assistant/issues/354 for that.

bocekm:
@AloisMahdal, this issue could arise before this fix in case:
- there was an executable file on the system which had non-UTF-8 coded name,
- the locale was set to UTF-8

If the locale is set to non-UTF-8 encoding then many more modules fail when reading any of the generated common files. So that's another issue.

AloisMahdal:
> I don't know what encoding has been used when saving the executable.log file.

AFAIK we create the file, so we do have control over its encoding (although we probably just leave it up to locale).

The problem is that if you have data and don't know the encoding, there's nothing you can do with it (no matter what chardet or other heuristics will tell you---given enough small sample, "unusual" content or similar encoding, they *will* be wrong).

Eg. what if there is sequence `11011011 10110111`?  Is that UTF-8? Sure, it can be a 2-byte valid UTF-8 sequence but also 2 valid characters from a plethora of extended-ASCII encodings.  Or even UTF-16? If you don't know, you don't know, right? Just because python's codec does not throw exception does not mean that the data is interpreted correctly, it's actually only subset of sequences that is *known* to be incorrect (ie. non-utf8).

So what is actually supposed to happen if in the situation as you described? Let's say we try to 
 open executable.log and there's non-utf-8 sequence, eg. `11011110 01110010`; what does preupgrade-assistant do after this PR?  IMO all it can do is fail (or throw it away, which is hardly a good strategy for upgrades).

bocekm:
> AFAIK we create the file, so we do have control over its encoding (although we probably just leave it up to locale).
Yes, we leave it up to locale.

Now, the Preupgrade Assistant (framework) does not read the `executable.log` file anymore. It just provides its path to modules. So it's up to a module how it handles the chars. Currently, they do not fail upon reading such content. These two modules using are the file:
1) system/hal
```
with open(VALUE_EXECUTABLES, "r") as f:
    filelist = f.readlines()
```
2) system/BinariesRebuild
```
do
  # some bash magic
done < "$CONTENT_OF_VALUE_EXECUTABLES"

**Comment by bocekm** [comment]
> IMO all it can do is fail (or throw it away, which is hardly a good strategy for upgrades).

I wouldn't vote for failing. Why shouldn't we let users have non-utf-8 filenames?

By the way, as I wrote above, the Preupgrade Assistant expects all the following generated common files to be utf-8 encoded, otherwise the module calling the python API function using this file will fail:
VALUE_RPM_QA
VALUE_CHKCONFIG
VALUE_CONFIGCHANGED
VALUE_RPM_RHSIGNED
To fix that I created #354.

**Comment by AloisMahdal** [comment]
> [...] Now, the Preupgrade Assistant (framework) does not read the executable.log file anymore [...]

...

> [...] Preupgrade Assistant expects all the following generated common files to be utf-8 encoded [...]

Wait, which one is it?

If we (framework) provide a file, we must also provide its format specification (encoding being just part of that), and expect both parties to commit to that specification.  Now whether the specification dictates a particular encoding or just relies on locale being set, then it has to apply to both sides: whoever creates that file *and* who reads it.  If this is not the case, there's no way of saying who is wrong.


> I wouldn't vote for failing. Why shouldn't we let users have non-utf-8 filenames?

It's not that we don't allow them, it's that by calling `line.decode('utf8')' you expect `line` to be utf-8. If that is not the case the decoder will fail (*if* you are lucky enough).  There's just no good way to deal with that.

**Comment by bocekm** [comment]
> > [...] Preupgrade Assistant expects all the following generated common files to be utf-8 encoded [...]
> Wait, which one is it?

These are generated by commands in `/usr/share/preupgrade/data/preassessment/scripts.txt`
```
# ls /var/cache/preupgrade/common/
allmyfiles.log  executable.log  passwd.log      rpm_qa.log        rpmtrackedfiles.log
chkconfig.log   group.log       rpm_etc_Va.log  rpm_rhsigned.log  rpm_Va.log
```

> If we (framework) provide a file, we must also provide its format specification (encoding being just part of that), and expect both parties to commit to that specification. Now whether the specification dictates a particular encoding or just relies on locale being set, then it has to apply to both sides: whoever creates that file and who reads it. If this is not the case, there's no way of saying who is wrong.

This comment does not seem relevant to this PR. Now the filenames on the system can be encoded with whatever encoding because the PA does not use the `executable.log` anymore. Your comment seems more relevant to #354.

bocekm:
@AloisMahdal, please come and visit me and I can explain you what is this two line change is supposed to fix. Then you can decide whether it's something worth creating a test for.

bocekm:
@AloisMahdal stopped by and agreed to not write tests for this PR.

Issue #322, opened 2017-10-10
Support RHEL 6.6 (RHEL 6.7+ is supported now)


bocekm:
**Fix the error:**
```
File "/usr/lib/python2.6/site-packages/preupg/utils.py", line 537, in config_has_section
  config = configparser.RawConfigParser(allow_no_value=True)
TypeError: __init__() got an unexpected keyword argument 'allow_no_value'
```

**Update the Requires in spec to:**
```
openscap%{?_isa} >= 0:1.0.8
openscap-engine-sce%{?_isa} >= 0:1.0.8
openscap-utils%{?_isa} >= 0:1.0.8
```

**Get rid of python-six**
python-six is not provided on RHEL 6.6

bocekm:
@pirat89, I wonder why the error does not show up also on RHEL 6.7+, because per [doc](https://docs.python.org/2/library/configparser.html) the `allow_no_value` was added in python 2.7. And even RHEL 6.9 has still python 2.6.6.

pirat89:
@mbocek: because of downstream patches

pirat89:
@bocekm: we need to resolve troubles with pykickstart, which is not part of RHEL 6.6. We can bundle it (so functionality will be kept for RHEL 6.6). OTOH we could just separate the code to a subpackage, so users will be able to do assessment of the system and get report file. After upgrade to newer version of RHEL 6, they could install the subpackage when they want it to use it (it is partial and messy solution, but it could be fast).

bocekm:
Resolved by https://github.com/upgrades-migrations/preupgrade-assistant/pull/359

PR #355, opened 2018-02-19
Get rid of python-six

pirat89:
Remove last use of the python-six library as we do not plan to support Python 3 and it breaks dependencies for RHEL 6.6.

Related: #322


pirat89:
I am closing this PR as it will be easier to have it together with PR that resolves support of RHEL 6.6.

PR #356, opened 2018-02-19
WIP: Support RHEL 6.6

### LINKED ISSUE #322: Support RHEL 6.6 (RHEL 6.7+ is supported now)

pirat89:
PA should support RHEL 6.6 for assessment of the system. Current state:

- [x] get rid of python-six (missing on RHEL 6.6)
- [x] update requirements in spec
- [x] resolve troubles with RawConfigParser (see #322)

Resolves #322


pirat89:
Note: I was thinking about method is `SystemIdentification` clast like that:
```python
    @staticmethod
    def get_rhel_version():
        """Returns tuple with major and minor version of RHEL or None."""
        if SystemIdentification.get_variant() is None:
            return None
        redhat_release = FileHelper.get_file_content("/etc/redhat-release", "rb")
        return tuple([int(x) for x in redhat_release.split()[6].split(".")])

```

But to be used already at least once, the class had to be indeprendent on `FileHelper` and has to be defined before `configparser` is imported in the file. For now, I used solution that you can see in the third commit.

pirat89:
@AloisMahdal: Lojzo, for testing purposes:
- should be possible to install preupgrade-assistant without broken dependencies
- assessment is possible on RHEL 6.6 without errors generated by PA itself.

AloisMahdal:
`qa_ack`: Just run preupgrade-assistant test suite on RHEL-6.6 (`ipu tier pa=P356 pa-* > p356.sh; edit p356.sh; sh p356.sh`).  (Watch for, and fix any incompatibilities caused by test suite itself.)

AloisMahdal:
> Watch for, and fix any incompatibilities caused by test suite itself

for example, old openscap merges stdout and stderr from `check` script, so `preupgrade_assistant__run1` will trigger "stray output" false-positive with any risk.  IOW, we need to update test library prior to any testing.

bertinatto:
@AloisMahdal, after discussing a96c9fe with @bocekm and @mfranczy, we decided to try an alternative approach. This is reflected on PR #359.

I don't seem to be able to set the correct tags there, so I'll ask you here: is it possible to perform the checks of that PR instead of this one?

Thanks.

AloisMahdal:
Acked the other one.  Since both PRs are doing the same change from user POV, we can test the other one with exactly the same test procedure.

bocekm:
This PR has been obsoleted by https://github.com/upgrades-migrations/preupgrade-assistant/pull/359.

PR #358, opened 2018-02-21
Keep %description wrapped to ~80 characters

AloisMahdal:
Unwrapped, the text will fold unpredictably (ie. in the middle of a wo
rd,) at any terminal width.


bocekm:
I've seen that today too when creating a downstream build and it indeed looked bad.

PR #359, opened 2018-03-06
Support RHEL 6.6

bertinatto:
This workd is based on #356, where instead of bundling `ConfigParser` it removes the legacy config options that required the use of `allow_no_value`.

### DIFF
--- a/etc/preupgrade-assistant.conf
+++ b/etc/preupgrade-assistant.conf
@@ -2,12 +2,3 @@
 home_directory_file=.preupgrade_dirs
 # Syntax is mentioned below in section [home-dirs]
 user_config_file=enabled
-
-[home-dirs]
-# User is responsible for valid input in this part.
-# All paths should be escaped if contains e.g. whitespaces
-# Otherwise directories/files can be skipped.
-# if you want to add directories items which start with "#" or looks likes
-# config section, supported format is ./#name or ./[something]
-.config/
-.mozilla/
--- a/packaging/preupgrade-assistant.spec
+++ b/packaging/preupgrade-assistant.spec
@@ -38,7 +38,6 @@ BuildRequires:  python-setuptools
 BuildRequires:  rpm-python
 BuildRequires:  diffstat
 %if 0%{?rhel}
-BuildRequires:  python-six
 BuildRequires:  pykickstart
 %endif # RHEL
 Requires(post):   /sbin/ldconfig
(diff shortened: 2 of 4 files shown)


bertinatto:
@mfranczy, @jmazanek:

Please, can you take a look?

bertinatto:
@jmazanek, @bocekm:

I can't add the proper labels to this PR. Maybe this is a permission issue. Can you try it?

AloisMahdal:
`qa_ack`: Just run preupgrade-assistant test suite on RHEL-6.6 (`ipu tier pa=P356 'pa-*' > p356.sh; edit p356.sh; sh p356.sh`).  (Watch for, and fix any incompatibilities caused by test suite itself.)

For example, old openscap is buggy and merges stdout and stderr from `check` script, so `preupgrade_assistant__run1()` from the test library will trigger "stray output" false-positive with any risk.  IOW, we need to update test library prior to any testing.

Another thing to watch for is if `ipu` and supporting tasks will work properly, they have never been tested neither with RHEL-6.6 (i don't expect much of a difference, though).

(Note that this is alternative to #356, so the testing is the same.)

tcerna:
Hi,
currently it seems well. 

Old PA failed on RHEL-6.6:
```
# preupg --force --select-rules=xccdf_preupg_rule_system_SystemVersion_check 
__init__() got an unexpected keyword argument 'allow_no_value'
```

New version (applied patch from this PR) of PA:
```
# preupg --force --select-rules=xccdf_preupg_rule_system_SystemVersion_check 
Generating simply styled report due to the limitations of the installed OpenSCAP
001/001 ...done    (Checking the system version and variant) (time: 00:00s)
The assessment finished (time 00:00s)
Result table with checks and their results for 'main contents':
------------------------------------------------------
|Checking the system version and variant      |fail  |
------------------------------------------------------
The tarball with results is stored in '/root/preupgrade-results/preupg_results-180321093428.tar.gz' .
The latest assessment is stored in the '/root/preupgrade' directory.
Summary information:
We have found some critical issues. In-place upgrade or migration is not advised.
Read the file /root/preupgrade/result.html for more details.
Please ensure you have backed up your system and/or data 
before doing a system upgrade to prevent loss of data in 
case the upgrade fails and full re-install of the system 
from installation media is needed.
Upload results to UI by the command:
e.g. preupg -u http://example.com:8099/submit/ -r /root/preupgrade-results/preupg_results-180321093428.tar.gz .
```
We can see in result details that there is extreme risk result that current machine is not the latest RHEL-6.9.

Let's wait for results of test suite on this pull request.

bocekm:
@tcerna, yeah, the extreme risk is expected, as you used modules for RHEL 6.9->7.4 upgrade. But the preupg execution does not end up with traceback which used to do, so I would call it verified.
Please add qa_verified label to the PR and I'll merge it.

Issue #343, opened 2017-11-06
error messages when stdin is redirected from pipe


AloisMahdal:
Running

    preupg <<<y

will produce multiple copies (4 with simple passing mock module) of this error message on stderr:

    stty: standard input: Inappropriate ioctl for device

This has been [known for some time][1]; I've verified this still appears with latest master (`2.5.0-1.201711021818Z`).

  [1]: https://bugzilla.redhat.com/show_bug.cgi?id=1331416#c3

AloisMahdal:
@bocekm Could "we" make it in next milestone?  (We have it covered in tests for long time already.)   It keeps spamming our test logs so it would be nice to get rid of.

bocekm:
I'll schedule it for the next milestone. We'll see what the priorities will be. We would accept even your PR ;)

PR #362, opened 2018-05-11
Fix issue with redirected stdin

pirat89:
Ignore stderr when can't get width of terminal. The value is used only for formatting pusposes and default value is still 80.

The get_terminal_width method returns now really the width value. Before it returns only string with height and width.

Resolves: #343

### DIFF
--- a/preupg/scanning.py
+++ b/preupg/scanning.py
@@ -2,6 +2,7 @@
 from __future__ import unicode_literals
 import datetime
 import os
+import subprocess
 from preupg.logger import settings, logger_report, log_message, logging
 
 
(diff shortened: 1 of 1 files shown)


AloisMahdal:
`qa_ack`: already covered in downstream test suite; just port it to the [new public suite][1] and enable the relevant part.

   [1]: https://github.com/AloisMahdal/jats-preupgrade-assistant

pirat89:
Modified commit msg.

AloisMahdal:
Test migrated and ran with *preupgrade-assistant-2.6.0-0.201805111535Z.PR362.el6_8.noarch*.  Result is PASS.

PR #364, opened 2018-09-05
Bump version to 2.6.1

pirat89:
- fixes issues with redirected stdin
- provides paths to all significant directories in PA API


AloisMahdal:
LGTM

